PR #1, opened 2016-08-06
Add MIT/Apache2 License, Readme

killercup:
Let's get this out of the way :)

@colin-kiegel, this is of course just a suggestion. MIT/Apache-2.0 is what most Rust libraries, incl. custom_derive, use.


colin-kiegel:
perfect - thx.

Issue #3, opened 2016-08-06
CustomDefault


killercup:
The builder pattern allows us to set optional values on a struct—but how do we determine the default values for those fields? Rust's default answer (hehe) is of course [`std::default`](https://doc.rust-lang.org/std/default/index.html) with its `#[derive(Default)]` and custom-written `impl Default`.

At first, we can
- let the user add `#[derive(Default)]` so they can do `MyStruct::default().setter(42)`,
- or `impl MyStruct { pub fn new(…) {…} }` so they can do `MyStruct::new(…).setter(42)`.

It would be really nice to also offer a way to derive/generate a `Default` impl with custom values, e.g.:

``` rust
custom_derive! {
    #[derive(CustomDefault)]
    struct Channel {
        id: Uuid,
        token: Authentication,
        #[default(42i32)] // ← custom attribute
        special_info: i32,
    }
}
```

(How) should we do this? (I don't think it's necessary to have this early on, and it should probably be a separate crate.)

@colin-kiegel already suggestion this in https://github.com/DanielKeep/rust-custom-derive/issues/18.

colin-kiegel:
Since `derive(Builder)` aims to improve convenience, we should be looking for a simple way for the user to specify 'custom defaults'.

I think there are three approaches
1. incorporate customizing defaults of optional fields into `derive(Builder)`
2. require a Default implementation for optional fields (not necessarily the struct)
3. require a Default implementation for the struct (not necessarily its items)

**Option 1**:
* +: We can avoid a `Default`-implementation of required fields.
* +: Looks convenient to use to me.
* -: But this increases our problem scope significantly.
=> We may postpone it to a version 2.x, or something like that - at least to get started.

**Option 2**:
* +: We can avoid a `Default`-implementation of required fields.
* -: Definitely not convenient to use, right? (imagine a struct with multiple fields of the same type, but different defaults - new-type pattern + much extra work? - ugh!).

**Option 3**:
* +: Convenient for the user
* +: Convenient for us to begin with (smaller problem scope).
* -: User can circumvent required fields.

By the principle of exclusion, I'd say lets start with **option 3**, i.e. let `derive(Builder)` require a `Default`-implementation for the same struct it acts on (for at least versions 0.x - 1.x)

This artificial restriction gives us some _legroom_ (=flexibility):
* Our perspective: Specifying defaults will not be part of our problem-scope
[...]

PR #4, opened 2016-08-06
Add first impl

killercup:
Sorry, I couldn't stop myself! 😄

Best review by reading the commits one after another.


colin-kiegel:
wow, that was fast - I will need some time for that. :-)

killercup:
I think it makes sense to land a pretty basic version and add stuff one feature at a time. This has a few tests (but not nearly enough) and just implements setters.

I need to investigate how to best generate getter names. I really want this to generate idiomatic Rust, so getters names would be prefixed with `get_`, e.g. `get_foo()` and `get_foo_bar()`.

If I read <https://github.com/rust-lang/rust/issues/29599> correctly, it's not possible to generate `ident` on stable. If we were to do getters on nightly only, [interpolate_idents](https://github.com/SkylerLipthay/interpolate_idents) might be helpful ([usage example](https://github.com/aidanhs/ayzim/blob/713497c264/src/cashew.rs#L162-L178)).

A more creative solution for stable might be creating a single `get(&self) -> new_sub_module::Getter` method, where `Getter` is a newtype that implements methods to get each of the struct fields, so you can do `x.get().foo()`.

killercup:
Btw, feel free to enable travis and configure it some more (to render docs and stuff) – I just added a basic file so we don't forget about it :)

src/derive_builder.rs
```diff
@@ -0,0 +1,158 @@
...
+/// fn main() {
+///     let x = GenLorem::default().ipsum("sit").dolor(42);
+///     assert_eq!(x, GenLorem { ipsum: "sit".into(), dolor: 42 });
+/// }
+/// ```
+///
+/// ## Gotchas
+///
+/// - Tuple structs and unit structs are not supported as they have no field
+///   names.
```
colin-kiegel: That's ok for me. I think it's not worth the effort to try to support them - because some kind of implicit naming is needed after all. So let's just require explicitness here. :-)

src/derive_builder.rs
```diff
@@ -0,0 +1,158 @@
...
+/// - When defining a generic struct, you cannot use `V` as a generic parameter
+///   as this is what all setters are using.
```
colin-kiegel: I sugest to use something with even less collision probability: https://www.random.org/strings/?num=30&len=5&upperalpha=on&unique=on&format=html&rnd=new or maybe our initials PHCK or some other pseudo-random acronym with 4+ characters?

src/derive_builder.rs
```diff
@@ -0,0 +1,158 @@
...
+        ),
+        fields = [$({
+            field_name: $field_name:ident,
+            column_name: $column_name:ident,
+            field_ty: $field_ty:ty,
+            field_kind: $field_kind:ident,
+        })+],
+    ) => {
+        #[allow(dead_code)]
+        impl<$($generics),*> $struct_name<$($generics),*> {
```
colin-kiegel: shall we already _reserve_ an implementation of `new()` which delegates to `Default::default()` as long as we don't yet support required fields? The same goes for the non-generic struct case of course.

src/parse_struct.rs
```diff
@@ -0,0 +1,298 @@
...
+///     }, {
+///         field_name: baz,
+///         column_name: other,
+///         field_ty: String,
+///         field_kind: regular,
+///     }],
+/// }
+#[macro_export]
+#[doc(hidden)]
+macro_rules! __diesel_parse_struct_body {
```
colin-kiegel: cool stuff - how do you suggest to proceed with this?
a) copy-paste this from diesel _unchanged_ as needed
b) minify it to our needs
c) ...?

tests/basic_setters.rs
```diff
@@ -0,0 +1,38 @@
...
+    struct Lorem {
+        ipsum: String,
+        dolor: String,
+        sit: i32,
+        amet: bool,
+    }
+}
+
+impl Lorem {
+    pub fn new<T: Into<String>>(value: T) -> Self {
```
colin-kiegel: Oh, what a day, I didn't know multiple `impl Item`s are possible. Guess I never tried. :-)

But this would collide, if we reserved `fn new()` for ourselves.

colin-kiegel:
looks awesome. :-)

colin-kiegel:
PS: added travis.yml in master - please feel free to finetune it. I just copied a template from another project...
* E.g. coveralls probably does not make sense for a macro crate.
* #6

src/derive_builder.rs
killercup: Good point. I think this will end up in the documentation, so I wanted to use a "regular" generic name. How about VAL or VALUE? No need to pick some obscure letters :)

src/derive_builder.rs
killercup: I think a `new` method is something users should explicitly write themselves. We don't have enough information to to anything good anyway (right now).

src/parse_struct.rs
killercup: I just copied it because I knew it would do the job and is nicely documented. :)

> b) minify it to our needs

sounds good

> c) ...?

c) Publish a crate that exposes this macro exactly

tests/basic_setters.rs
killercup: Yeah, you can write all the `impl`s you want :)

> But this would collide, if we reserved fn new() for ourselves.

Yep. Which is why I don't want us to write a `fn new`. This way we don't need to deal with required fields and people can always just do their own thing. (And maybe later switch to a new version of this crate that does more stuff for them.)

killercup:
Let me just rebase this on master to make it mergeable again.

src/derive_builder.rs
colin-kiegel: ok, VALUE is fine, since it's not CamelCase.

src/derive_builder.rs
colin-kiegel: hm - but don't you think we will be generating meaningful `new`-methods sooner or later?

killercup:
I added `TRAVIS_CARGO_NIGHTLY_FEATURE=""`, by the way, so this actually has a green build now :)

> coveralls probably does not make sense for a macro crate.

I'm not sure. I might be clever enough to tell us that we are not testing all the code the macro is generating. Let's wait and see what it says.

src/derive_builder.rs
killercup: Actually, no—`CustomDefault` (#3) might, though! 😄

colin-kiegel:
Yippieh :-)

Btw: Do you prefer a squashed merge or the usual one in these cases?

killercup:
I prefer regular merges with all commits/history intact, since I try to build up features and it's nice to see how something was put together. When there are too many small fixes in response to feedback I'm cool with squashing those, though.

src/derive_builder.rs
colin-kiegel: ok, even better. It's beginning to all fall into places. Nice. :-)

.. and since it is _Custom_ Default, it may even be a custom method name: `$name_me(...) -> Foo`

src/derive_builder.rs
sgrif: Does it actually collide? Hygiene should protect you here.

killercup: I think `impl<T>` means `T` is declared for the entire `impl` and all its `fns`. [This code](https://play.rust-lang.org/?code=struct%20Foo%3CT%3E%20%7B%0A%20%20%20%20bar%3A%20T%2C%0A%7D%0A%0Aimpl%3CT%3E%20Foo%3CT%3E%20%7B%0A%20%20%20%20fn%20barify%3CT%3E(%26self%2C%20bar%3A%20T)%20%7B%0A%20%20%20%20%20%20%20%20unimplemented!()%3B%0A%20%20%20%20%7D%0A%7D&version=stable&backtrace=0) says

> type parameter `T` shadows another type parameter of the same name

sgrif: Right, but the `T` in your macro definition should not collide with `T` which was passed in as an ident. This is why we're able to do nonsense with the parse macro to write code like `let (x, x, x, x, x, x, x, x) = self` but have each `x` be a different variable.

colin-kiegel: just confirmed, it **does** collide, although the TypeParameters come from different hygiene contexts. We could file this as a bug upstream.

sgrif: Hey man, we were having a speculation party over here and you had to go and ruin it by actually **seeing** what it does?

killercup: @sgrif, that was my expectation as well.

> just confirmed, it does collide

> ruin it by actually seeing what it does?

Lol, you're all crazy. Compile-fail test or it didn't happen! 😆

killercup: But seriously, changing the `<T>` in [this line](https://github.com/colin-kiegel/rust-derive-builder/blob/0cba6c64147019d4f48a168dce45e8437022f18e/tests/generic_structs.rs#L6) (and further down) to `VALUE` does indeed error with

```
error[E0194]: type parameter `VALUE` shadows another type parameter of the same name
  --> <derive_builder macros>:15:1
```

Feel free to report this to anyone who might want to fix it :)

Issue #13, opened 2016-08-07
Support custom documentation for setters


mre:
Consider this builder:
```rust
custom_derive! {
    #[derive(Debug, Default, Builder)]
    struct Channel {
        //! This is a comment for token
        token: i32,
        special_info: i32
    }
}
```
It would be great if there was a way to add the comments as documentation for the fields:

```rust
//! This is a comment for token
fn token<VALUE: Into<i32>>(mut self, value: VALUE) -> Self {
    self.token = value.into();
    self
}
```

No clue if that is possible, though. :smile:

killercup:
This should be possible, `/// hi there` comments become `#[doc(hi there)]` attributes.

Would you want to differentiate between comments for the struct field and comments for the setter? (If the field is not public, its docs wouldn't be rendered, so it might be cool to just add them to the setter fn.)

mre:
Thanks for the quick response.
Documentation for the setter would be enough for me right now. But a fallback for non-public fields would be totally awesome. Depends on how tricky it is to implement.

colin-kiegel:
@mre thank you very much for your feedback. :-)

A simple solution would be to allow item doc-comments, but to skip over them and use auto-generated documentation for setters. Spontaneously that is what I would try first and then wait and see if there is a big need for something more sophisticated or magic.

colin-kiegel:
Ok, for now we have auto-generated doc-comments (https://github.com/colin-kiegel/rust-derive-builder/commit/95f3b4b0a74a42a24393c4fd93cf53afe9fbddcf)
```rust
/// Set `$field_name`
///
/// auto-generated by [derive_builder](https://crates.io/crates/derive_builder)
```

Skipping doc-comments (and `pub` keywords) in the struct definition is a bit more delicate and therefore a pull request https://github.com/colin-kiegel/rust-derive-builder/pull/14

I am open for future changes - but I think this is at least an improvement of the current state. I will release now without the pull request. ;-)

colin-kiegel:
implemented and published with [v0.2.0 (Changelog)](https://github.com/colin-kiegel/rust-derive-builder/blob/master/CHANGELOG.md#020---2016-08-22)

PR #14, opened 2016-08-07
parse doc comments and visibiliy specifier

colin-kiegel:
implementation according to https://github.com/colin-kiegel/rust-derive-builder/issues/13#issuecomment-238099570

@killercup Seems to work just fine, but it's a delicate / complex macro. At leas a review seems adequate - and we haven't really decided yet.


colin-kiegel:
PS: current behaviour is 'unexpected token', which could be classified as a bug and is fixed by this PR.

killercup:
Sorry, I'm a bit out of the loop on this—disregard anything that's totally
obvious once I properly look at this.

What are you aiming for with this? Where do you want to skip doc/vis? Can
you add some tests for this change? (Did we add macro-compile tests yet to
compare the expanded tokens? If not, we should probably do that. Should be
possible with static-cond.)


Colin Kiegel <notifications@github.com> schrieb am So., 7. Aug. 2016 um
23:27 Uhr:

colin-kiegel:
this allows public and/or annotated struct fields, by just skipping over
these tokens. The current parser errors with unexpected token in these
cases.

will add a test for this and I might try to combine these matchers to save
iterations. :-)

PS: I was thinking about testing expanded tokens too.

Pascal Hertleif <notifications@github.com> schrieb am Fr., 12. Aug. 2016,
14:56:

colin-kiegel:
Ok, I did this:
* rebased to master
* some additional refactoring / simplification
* and last but not least a new feature: **all** field attributes now map to setter functions

Still work in progress (obviously)
* Travis is failing on stable + beta (I tested only with nightly). Will have a look.
* I am still thinking we should _start_ with whitelisting specific attributes, to be on the safe side (IF that is possible).

PS: I splitted my changes into a history of rather small commits.

colin-kiegel:
I introduced whitelisting of attributes, only these will be added to setter-fns
* `#[doc = ...]`
* `#[cfg(...)]`
* `#[allow(...)]`

If there is demand, we could add these in the future, but I guess they probably make little sense in these positions
* `#[deny(...)]`
* `#[forbid(...)]`
* `#[warn(...)]`

All current attributes are listed here. IMO the rest of them doesn't make sense for setter-fns.
https://doc.rust-lang.org/reference.html#attributes

colin-kiegel:
PS: I also re-ordered most of the arguments to introduce `@foo` as first tokens in recursive calls (where foo = current_state_of_macro, e.g. `parse`).

colin-kiegel:
PPS: If we go along this road, next thing might be to parse and filter the struct-attributes and transfer them to the impl too (candidates again `doc`+`allow`, but `cfg` probably not needed). However I wonder if parsing struct-attributes would survive a possible transition to procedural macros 1.1. Therefore we might wait until then before we do it.

colin-kiegel:
@killercup undid the rebase which caused the trouble - ready for review

PR #17, opened 2016-08-08
make setter-methods public

colin-kiegel:
@killercup don't we want to make setter-methods public by default? E.g. the use case of @mre is to write a public api - possibly with builder objects.

We could introduce an opt-in or opt-out setting later.


killercup:
Oh sorry, my mistake. These should've been public.

Issue #19, opened 2016-08-13
more tests


colin-kiegel:
- [ ] expanded tokens
- [x] nice-to-have: compile fail tests with compiletest
- [ ] nice-to-have: absence of clone in complex scenarios -- if we switch to `&mut self` signatures
  ...more?

colin-kiegel:
I'm quite happy with the current tests now. :-)

We don't test full expansion, but a combination of unit tests on code generation and behavioural tests instead. That seems better to me.

PR #20, opened 2016-08-13
Feature/non consuming

colin-kiegel:
This PR is based on #14 .. assuming, we merge in this order. Docs and changelog are up-to-date.

Before we merge this one, I would verify (or at least understand) whether the optimization promise survives the multi-crate case.

If we merge, I think we are good to go for 0.2.0. :-)


killercup:
Wow, you were quite busy, @colin-kiegel!

I guess you didn't mean to include the `*.orig`?

What's up with travis? There seem to be problems on stable and beta with parsing the generics code:

```plain
<derive_builder macros>:10:38: 10:48 error: expected identifier, found `'a`
<derive_builder macros>:10 # [ allow ( dead_code ) ] impl < $ ( $ generics ) , * > $ struct_name < $ (
                                                                ^~~~~~~~~~
<derive_builder macros>:10:38: 10:48 error: expected identifier, found `T`
<derive_builder macros>:10 # [ allow ( dead_code ) ] impl < $ ( $ generics ) , * > $ struct_name < $ (
```

How do you propose we test this? I don't feel comfortable with releasing/merging without a good testing story.

colin-kiegel:
this is the offending commit - it's in master and I rebased and pushed everything without proper testing on all rust versions .. :-/
* https://github.com/colin-kiegel/rust-derive-builder/commit/fda318a9adc447396af31378d89800f13a2296cd

Off-topic: To avoid future problems like this I invested some time in a proper pre-commit git hook which will force me to run all tests on _stable+beta+nightly_. Everything not staged for the commit will be stashed away before running the tests.

When I am home this evening I will begin the repairs in master and this PR.
Btw. I will not do rebases in master, but only in PRs.

killercup:
I personally try to use @homu for all projects and never push to master myself. Sadly, homu.io seems to be no longer maintained. @badboy runs his own instance as @not-a-robot, so if you ask nicely…

> Btw. I will not do rebases in master, but only in PRs.

You can restrict that in Github as well.

colin-kiegel:
@killercup undid the rebase which caused the trouble - ready for review

This differs from #14 only in the last commit.

badboy:
Chiming in: I could add additional repos to my homu instance, though currently it requires an account with write access to the repo (in effect: my account or the bot's account). That is only for the setup though. So if you want that, add one of the accounts and ping me.

colin-kiegel:
@badboy: Thx - I added your account. Does it work by sending commands to your bot then? Is there a list of valid commands somewhere? Who is authorized to issue commands? ... If someone could point me to something like a readme, that would be great. :-)

@killercup ok .. one final rebase to restore mergebility

PS: I've made my pre-commit a pre-push hook -- looks nice, has colored output, expands the output if an error occured.. Once I'm done tweaking, I'll share this too. :-)
```
$ git commit # or push
II: Stashing any unstaged changes... ✓ skipped
II: Running tests on stable... ✓ 
II: Running tests on beta... ✓ 
II: Running tests on nightly... ✓ 
II: Unstashing the stashed changes... ✓ skipped
pre-commit status: OK
```

tests/empty_struct.rs
```diff
@@ -0,0 +1,15 @@
+#[macro_use] extern crate custom_derive;
+#[macro_use] extern crate derive_builder;
+
+custom_derive!{
+    /// This is a doc comment for the struct
+    #[warn(missing_docs)]
+    #[allow(non_snake_case)]
+    #[derive(Debug, PartialEq, Default, Builder)]
+    struct IgnoreEmptyStruct {  }
```
killercup: FYI: The other empty struct syntax (`struct IgnoreEmptyStruct;`) does not work. Do we want to fix this? It's basically just another no-op case in the macro.

tests/basic_setters.rs
```diff
@@ -24,15 +24,15 @@ impl Lorem {
...
 #[test]
 fn setters() {
-    let x = Lorem::new("lorem")
-        .dolor("dolor")
-        .sit(42)
-        .amet(true);
+    let mut x = Lorem::new("lorem");
+    x.dolor(Some("dolor".into()));
+    x.sit(42);
+    x.amet(true);
```
killercup: Why did you change this to no longer user chaining? I think the chained version looks much more pretty. Just append `.clone()` (or add a method `build` next to `new` that calls `clone`).

src/lib.rs
```diff
@@ -1,49 +1,105 @@
...
 //! }
 //!
 //! fn main() {
-//!     let x = GenLorem::default().ipsum("sit").dolor(42);
+//!     let x = GenLorem::default().ipsum("sit").dolor(42).clone();
 //!     assert_eq!(x, GenLorem { ipsum: "sit".into(), dolor: 42 });
 //! }
 //! ```
 //!
+//! ## Doc-Comments and Attributes
```
killercup: Nice docs :)

killercup:
LGTM so far, with a few comments left inline.

Anything new on the testing front? Should we do token comparisons, and maybe also add compile fail tests with compiletest?

tests/empty_struct.rs
colin-kiegel: If we do want to support that, the bigger picture might be "don't fail on anything we are (a) not interested in and which is (b) parse-able by custom_derive". That would also include tuple-structs and enums. If you like that idea, we can open a low-priority tracking issue for that. :-)

tests/basic_setters.rs
colin-kiegel: Funny, if there was a reason I have forgotten it by now. I'll change it back. :-)

colin-kiegel:
Hooray. :-) Thx for your feedback. I'll switch the setter-test back to the chained-syntax and merge tomorrow. IMO this would then be a good point to package a new release (0.2 due to the changed setters) and make some advertisement in the user-forum and maybe reddit.

I have added a tracking issue for "more tests": https://github.com/colin-kiegel/rust-derive-builder/issues/19 Yes, I think we should do these kinds of tests if it doesn't involve arcane wizardry. I haven't tried yet. :-)

tests/basic_setters.rs
colin-kiegel: It occurred to me again. The Alternative would be an additional `.clone()` and `derive(Clone)` in order to bind to x, because we dont have a terminal build method. It just looked a bit weird to me, but it's ok. ;-)

PR #28, opened 2016-10-11
Macros one dot one (Do Not Merge)

badboy:
So today I [read up on the new Macros 1.1 approach](http://www.ncameron.org/blog/macros-and-syntax-extensions-and-compiler-plugins-where-are-we-at/) and it turns out it is really easy to use (together with the `syn` and `quote` crates).

I ported over the code to use Macros 1.1 and it works with significantly less code now.
I had to move the tests into their own crate, compiling a test binary in the right library mode is impossible.

**This makes the code require a recent nightly, DO NOT MERGE immediately**

This can probably stay around until Macros 1.1 are actually stable (probably not before the end of the year though).


killercup:
Woohoo! But I _really_ want to merge this right now! 😅

It might make sense to release this as an alternative version (if it's any good, I haven't read your code yet!). The macro derive code should be semi-stable in nightly -- diesel 0.8 pretty much promises you can use the latest nightly/syntex and it will work.

Hw much work will testing two crates be? I.e., we have macro_rules & macros1.1 and a set of integration tests that work for both.

badboy:
After I pushed this I thought about having both code parts behind feature flags. But then again, this seems less optimal. We could however make it a derive-builder-next package to be released now and merge both back when Macros 1.1 hit stable.

colin-kiegel:
I would like to talk with you guys about renaming things to `derive(setter,
getter)` and making `derive(builder)` a meta crate which only wraps things
with the _option_ to depcrecate in the future depending on usage. But I
expect the alternative name much more intuitive.

I had this thought for quite a while and I think this is a good moment to
actually do it. What do you think?

Jan-Erik Rediger <notifications@github.com> schrieb am Di., 11. Okt. 2016,
19:54:

src/lib.rs
```diff
@@ -20,22 +19,16 @@
...
 //! [builder pattern]: https://aturon.github.io/ownership/builders.html
-//! [custom_derive]: https://crates.io/crates/custom_derive
 //!
 //! # Examples
 //!
-//! This crate is best used in combination with [custom_derive].
-//!
 //! ```rust
-//! #[macro_use] extern crate custom_derive;
 //! #[macro_use] extern crate derive_builder;
```
badboy: I need to try.

src/lib.rs
```diff
@@ -107,5 +94,47 @@
...
+extern crate proc_macro;
+extern crate syn;
+#[macro_use]
+extern crate quote;
+
+use proc_macro::TokenStream;
+
+#[proc_macro_derive(Builder)]
+pub fn derive(input: TokenStream) -> TokenStream {
+    let input: String = input.to_string();
```
badboy: Indeed. that bit was copy'n'pasted

colin-kiegel:
PS: if we go down this road we should add options to support different setter styles

killercup:
@badboy instead of a new crate with a new name we could also do a "derive-builder v0.3.0.alpha-1" release or something like that (if @colin-kiegel wants to do that).

@colin-kiegel Interesting! I'd use the plural names though: `#[derive(Getters, Setters)]`

It might also be valuable to introduce a whole new level of 'indirect builder': By defining a new struct `FooBuilder` with the same fields as `Foo` but _only_ the builder methods (incl. `finalize`/`build`/`watshouldwecallit`) we can introduce a bit more type safety by making it very explicit when the building step is complete.

colin-kiegel:
@killercup Good points. :-)
cc @badboy

If you both like the idea of `#derive(Getters, Setters)` AND publishing as new crates `derive_setters` and `derive_getters` or just `setters` and `getters` we should act **fast**.. :-)

You may feel free to secure one set of crate names you like by publishing dummy crates. Because I'd guess it may be a popular exercise to implement this kind of stuff right now.

src/lib.rs
```diff
@@ -107,5 +94,47 @@
...
+        let f_name = &f.ident;
+        let ty = &f.ty;
+        quote!(pub fn #f_name<VALUE: Into<#ty>>(&mut self, value: VALUE) -> &mut Self {
+            self.#f_name = value.into();
+            self
+        })
+    });
+
+    quote! {
+        impl #impl_generics #name #ty_generics #where_clause {
```
dtolnay: Technical details which in no way affect my preference about where to put spaces:

- From the `quote!` macro's perspective there is no way for it to tell where you put spaces. It sees a `Vec<TokenTree>` which looks like `["impl", "#", "impl_generics", "#", "name", "#", "ty_generics", "#", "where_clause", Braces(["#", Parens(["funcs"]), "*"])]`.
- The string constructed by the `quote!` macro will have a single space in between every token (at least as of quote 0.3.3; I think in this version there might be more than one in certain cases). This is the simplest heuristic that is always correct with respect to rustc being able to parse the result as intended. So if `impl_generics` is `<T>` then the string will start with `"impl < T >".
- The builtin `stringify!` has the same behavior for the same reason:

    ```rust
    fn main() {
        // impl # impl_generics # name { # ( funcs ) * }
        println!("{}", stringify!(impl #impl_generics #name { #(funcs)* }));
    }
    ```

- When you take the string from `quote!` and parse it to a TokenStream for #[derive(Builder)] to return, the spaces are throw out again. The TokenStream looks like `["impl", "<", "T", ">", ...]`.
- One of the first things Rust does upon getting back the TokenStream is parse it. So if you [`cargo expand`](https://github.com/dtolnay/cargo-expand) a crate containing #[derive(Builder)] the spaces will be where Rust's pretty-printer puts them, not where `quote!` puts them and not where you put them.

Nontechnical details which fully determine my preference:

- I find `impl#impl_generics #name#ty_generics` less readable.
- I prefer the style @badboy used here.
- He probably got it from my examples so it is no coincidence :stuck_out_tongue:.

Bottom line:

- Put spaces wherever you want.
- There are so many steps between where you put the spaces and where the spaces end up that even if most of them mattered, it still wouldn't matter.

colin-kiegel:
There seems to be a way to use the `syn` crate in a build script for custom derives:
* https://github.com/dtolnay/syn#custom-derives-on-stable-rust

I'm not sure if it's worth investigating, since macros 1.1 should not be too far away now.

killercup:
Any plan on finishing/merging this? 1.15 with stable macros 1.1 will be released on February 3rd, and I'd be delighted to ship a new, shiny derive-builder version on day one :)

colin-kiegel:
Good point. I'll try to look into updating this PR this weekend.

PR #32, opened 2017-01-14
Feature/macros ono dot one

colin-kiegel:
@killercup: Here is the current state of the `macros_one_dot_one` branch. I continued the work of @badboy from #28.

I added backward compatibility to our existing feature:
* allow attributes on fields

I also introduced new features:
* select different setter patterns (instead of always mutable), e.g. `#[setters(immutable)]`
* select visibility of setters (public/private), e.g. `#[setters(private)]`
* logging via env_logger for better debugging, e.g. `RUST_LOG=derive_builder=trace cargo test`

Things I am not yet quite happy with
* Parsing of attributes: The attribute representation by the syn crate is already very complex. In addition to that our derive_builder crate has quite a bunch of options which I expect to even grow in the future. This combination leads to quite nested attribute parsing in `src/options.rs`. I wonder if there is a cleaner approach to it.
* I think it's time to add compilation tests or to test the generated tokens (e.g. private setters have a lacking testcase right now)
* There are some early but unfinished stubs for getters. I left them in for now, but I suggest to split this into a separate PR (or just implement it). But before I do that split, I would like to get some overall feedback first. :-)

Any suggestions are welcome.


colin-kiegel:
Note: Since all tests run inside of a sub-crate right now, we currently don't have doc-tests anymore. :-/

README.md
```diff
@@ -4,23 +4,22 @@
...
 
-**This is a work in progress.** Use it at your own risk.
+**This is a work in progress.** Use it at your own risk.  
+**This currently requires Rust nightly (2016-11-19 or newer), due to the usage of Macros 1.1.**
 
 And this is how it works:
 
 ```rust
-#[macro_use] extern crate custom_derive;
+#![feature(proc_macro)]
```
killercup: Don't need this no more!

src/options.rs
```diff
@@ -0,0 +1,284 @@
...
+        for attr in attributes {
+            ob.parse_attribute(attr);
+        }
+
+        ob.into()
+    }
+}
+
+#[derive(Default, Debug)]
+pub struct OptionsBuilder {
```
killercup: Heh, this code looks like a great indicator for a new "indirect builder" setting!

(Indirect builders are the "original" builder pattern using a struct with the same fields but all are optional.)

src/options.rs
```diff
@@ -0,0 +1,284 @@
...
+            setter_public: b.setter_public.unwrap_or(true),
+            setter_implicit_options: b.setter_implicit_options.unwrap_or(false),
+            getter_enabled: b.getter_enabled.unwrap_or(false),
+            getter_prefix: b.getter_prefix,
+            getter_public: b.getter_public.unwrap_or(true),
+        }
+    }
+}
+
+impl OptionsBuilder {
```
killercup: Seriously, the first bunch of methods can be derived, incl. the `warn!` output.

Also just as a warning, every tooling project I work on at some point ends up dogfooding itself :D

src/options.rs
```diff
@@ -0,0 +1,284 @@
...
+                    },
+                "getters" => {
+                        self.getter_enabled(true);
+                        self.parse_getters_options(nested_attrs);
+                        return
+                    },
+                _ => {}
+            }
+        }
+        debug!("Ignoring attribute.");
```
killercup: I would not use these attribute names, as they live in a global namespace and "getters" and "setters" is too generic (we just had that in diesel/rocket with "options"). How about something like `#[builder_options(setters(pattern = mutable, private = true), getter(public = true)]`?

src/options.rs
```diff
@@ -0,0 +1,284 @@
...
+            "public" => {
+                self.setter_public(true)
+            },
+            "private" => {
+                self.setter_public(false)
+            },
+            _ => {
+                panic!("Unknown option {:?}", ident)
+            }
+        };
```
killercup: I feel like there should be an idiomatic solution to parsing syn attributes to structs. Maybe another derive crate? A small serde plugin for (de)serializing things to/from rust attributes? The possibilities are infinite! :D

```rust
#[derive(Deserialize)] struct Options { setters: SetterOptions, /* … */ }
#[derive(Deserialize)] struct GetterOptions { pattern: SetterPattern, public: bool, private: bool }
#[derive(Deserialize)] enum SetterPattern { Owned, Mutable, Immutable }

// …
let attr: syn::Attribute = attr;
let opt: Options = parse_attribute!(builder_options);
```

src/options.rs
```diff
@@ -0,0 +1,284 @@
...
+        let value = parse_lit_as_cooked_string(lit);
+        match value.as_str() {
+            "implicit" => self.setter_implicit_options(true),
+            "explicit" => self.setter_implicit_options(true),
+            _ => panic!("Unknown value {:?}", value),
+        };
+    }
+
+    fn parse_getters_options(&mut self, nested: &[syn::NestedMetaItem]) -> &mut Self {
+        panic!("TODO: parse getter options -> {:?}", nested);
```
killercup: probably `s/panic!/unimplemented!`

killercup:
@colin-kiegel what about the doc tests [in lib.rs](https://github.com/colin-kiegel/rust-derive-builder/pull/32/files#diff-b4aea3e418ccdb71239b96952d9cddb6R25)? They can still do their thing, right?

colin-kiegel:
@killercup It looks like having `proc-macro = true` in `Cargo.toml` disables all doc-tests. I think it's a bug: https://github.com/rust-lang/rust/issues/39064

src/options.rs
colin-kiegel: Actually in published code I would like to get rid of these unimplemented functions. But generally you are right. :-)

src/options.rs
colin-kiegel: Yes, this kind of separation is, what I was looking for. :-)
Sounds interesting, I'll have a look at serde. Never used it before, but since everyone is talking about that's probably an education gap I should close anyway. :-)

src/options.rs
colin-kiegel: hm, I see what you mean.

I am open to a discussion. However I think we should try to keep it short, since the purpose of this crate is to avoid boiler plate. One of the next logical steps IMO would be to add options per struct field. I.e. prefix, pattern, visibility, etc. could all be changed on a per-field level. At least it should be possible to ignore individual fields - and if we introduce that, we can also introduce overwriting all the other options.

```rust
#[derive(Builder)]
#[builder_options(setters(pattern = mutable, private = true), getters(public = true)]
struct Foo {
   foo: u32,
   #[builder_options(setter(prefix = with, pattern = immutable, private = false), getter(prefix = has, public = false)]
   bar: u32
}
```

vs. 

```rust
#[derive(Builder)]
#[setters(mutable, private), getters(public)]
struct Foo {
   foo: u32,
   #[setter(prefix = with, immutable, public), getter(prefix = has, private)]
   bar: u32
}
```

src/options.rs
colin-kiegel: Yes, it felt like irony to write so much builder boilerplate in this crate. LOL. ^^

But how could we bootstrap if `derive_builder` depends on `derive_builder`? Shall we depend on the latest previously published version? But if that depends on the version before .. well .. people will notice strange dependency chains in their projects. :-)

killercup: Let's just adopt the new rustbuild build system! I can easiyl download a "stage0" snapshot of derive-builder before building derive-builder stage1, and then build-derive again with the stage1 build to finally get a stage2 version we can work with. :trollface:

colin-kiegel:
I updated the documentation - we can still change the syntax however.

src/options.rs
killercup: Good points, and I think you are right: Let's keep it concise, i.e. call the attributes `(s|g)etter(s?)`.

Another point: Do you think it'd be nicer to have setter and getter options defined in the same , or in two separate attributes?

killercup:
@colin-kiegel I suggest we merge this and iterate on it in further PRs

colin-kiegel:
ok, done. :-)

Btw: I've already started some refactoring to allow for field-based options. I'll try to wrap it up in a new PR later this evening.

killercup:
Nice, I'm looking forward to that.I'll also try to look into the builder varaints/typed builders again.

PR #34, opened 2017-01-19
prefixed setters and field specific overrides

colin-kiegel:
Ok, here is my prototype for
- field specific overrides
- prefixed setters

My approach was to replace `struct OptionsBuilder` with `struct OptionsBuilder<Mode: OptionsBuilderMode>`

Obviously there is a struct mode and field mode: https://github.com/colin-kiegel/rust-derive-builder/blob/feature/field-overrides/src/options.rs#L87-L107

I then do this to get field specific overrides on the options defined for the struct
https://github.com/colin-kiegel/rust-derive-builder/blob/feature/field-overrides/src/lib.rs#L271-L273

The approach is optimized for basically identical options on struct and fields. As soon as we will get options specific to one of these modes, we need to introduce fields and impls on the ModeStructs itself.

PS: I just realized that I didn't need the phantom data, I could just embed the Mode structs. LOL. ^^

### DIFF
--- a/src/options.rs
+++ b/src/options.rs
@@ -16,22 +16,14 @@ impl Default for SetterPattern {
 
 #[derive(Debug, Clone)]
 pub struct Options {
-    /// e.g. `#[setters]` (defaults to true)
+    // e.g. `#[setter]` (defaults to true)
     setter_enabled: bool,
-    /// e.g. `#[setters(owned)]` (defaults to mutable)
+    // e.g. `#[setter(owned)]` (defaults to mutable)
     setter_pattern: SetterPattern,
-    /// e.g. `#[setters(prefix="with")]` (defaults to None)
+    // e.g. `#[setter(prefix="with")]` (defaults to None)
     setter_prefix: String,
-    /// e.g. `#[setters(private)]` (defaults to public)
+    // e.g. `#[setter(private)]` (defaults to public)
     setter_public: bool,
-    /// e.g. `#[setters(options="implicit")]` (defaults to explicit)
-    setter_implicit_options: bool,
-    /// e.g. `#[getters]` (defaults to false)
-    getter_enabled: bool,
-    /// e.g. `#[getters(prefix="with")]` (defaults to None)
-    getter_prefix: String,
-    /// e.g. `#[getters(private)]` (defaults to public)
-    getter_public: bool,
 }
 
 impl Options {
@@ -54,22 +46,6 @@ impl Options {
     pub fn setter_prefix(&self) -> &str {
         &self.setter_prefix
     }
-
-    pub fn getter_enabled(&self) -> bool {
-        self.getter_enabled
-    }
-
-    pub fn _getter_visibility(&self) -> Option<quote::Tokens> {
-        if self.getter_public {
-            Some(quote!(pub))
-        } else {
-            None
-        }
-    }
-
-    pub fn _getter_prefix(&self) -> &str {
-        &self.getter_prefix
-    }
 }
 
 impl<'a, T> From<T> for Options where
@@ -84,27 +60,17 @@ impl<'a, T> From<T> for Options where
     }
 }
 
-pub trait OptionsBuilderMode {
-    // associated consts are not a thing yet :-/
-    // https://github.com/rust-lang/rust/issues/29646
-    fn setter_attribute_ident() -> &'static str;
-    fn getter_attribute_ident() -> &'static str;
-}
+pub trait OptionsBuilderMode {}
 
 #[derive(Default)]
 pub struct StructMode;
+
+impl OptionsBuilderMode for StructMode {}
+
 #[derive(Default)]
 pub struct FieldMode;
 
-impl OptionsBuilderMode for StructMode {
-    fn setter_attribute_ident() -> &'static str { "setters" }
-    fn getter_attribute_ident() -> &'static str { "getters" }
-}
-
-impl OptionsBuilderMode for FieldMode {
-    fn setter_attribute_ident() -> &'static str { "setter" }
-    fn getter_attribute_ident() -> &'static str { "getter" }
-}
+impl OptionsBuilderMode for FieldMode {}
 
 #[derive(Default, Debug)]
 pub struct OptionsBuilder<Mode: OptionsBuilderMode> {
@@ -112,11 +78,7 @@ pub struct OptionsBuilder<Mode: OptionsBuilderMode> {
     setter_pattern: Option<SetterPattern>,
     setter_prefix: Option<String>,
     setter_public: Option<bool>,
-    setter_implicit_options: Option<bool>,
-    getter_enabled: Option<bool>,
-    getter_prefix: Option<String>,
-    getter_public: Option<bool>,
-    phantom: ::std::marker::PhantomData<Mode>,
+    mode: Mode,
 }
 
 impl From<OptionsBuilder<StructMode>> for Options {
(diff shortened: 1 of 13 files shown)


colin-kiegel:
ok, I'm investigating the CI error. Looks like the `quote` crate introduced some breakage between 0.3.5 and 0.3.11.

colin-kiegel:
Ok, caused by this commit https://github.com/dtolnay/quote/commit/03f7b94e4549ac2cb47cb41bc40600f4979db1f3 :-)

src/options.rs
```diff
@@ -14,27 +14,31 @@ impl Default for SetterPattern {
...
     /// e.g. `#[setters(prefix="with")]` (defaults to None)
-    setter_prefix: Option<String>,
+    setter_prefix: String,
     /// e.g. `#[setters(private)]` (defaults to public)
     setter_public: bool,
     /// e.g. `#[setters(options="implicit")]` (defaults to explicit)
     setter_implicit_options: bool,
     /// e.g. `#[getters]` (defaults to false)
     getter_enabled: bool,
     /// e.g. `#[getters(prefix="with")]` (defaults to None)
```
killercup: I know this is not yet implemented, but to avoid confusion, I'd not use a different prefix example for getters, e.g. `get`. Also, I'd mention that this get's concatenated with the field name like `prefix_fieldname`.

src/options.rs
```diff
@@ -47,28 +51,63 @@ impl Options {
...
+        builder.parse_attributes(attributes);
 
-        ob.into()
+        builder.into()
     }
 }
 
+pub trait OptionsBuilderMode {
+    // associated consts are not a thing yet :-/
+    // https://github.com/rust-lang/rust/issues/29646
```
killercup: just for cross linking: https://github.com/rust-lang/rust/issues/29646

killercup: Also, I think you could use default impl for these fns

src/options.rs
colin-kiegel: +1 for explaining it in the documentation. Will do that.

This place here was not meant as a user documentation, but to help other contributors put the pieces together, if they are digging in our code. Should I switch to comments instead of doc-comments here?

Do you like that I automatically added the underscore as a separator, or would you prefer that users need to provide the full prefix including the underscore (more flexible, but more verbose)?

killercup: Oh right, users will need additional docs. I prefer doc comments for API-level comments, and regular double-slash comments for implementation details (that don't leak to the API level).

Keep the automatic underscore injection. It's what I'd expect.

src/options.rs
colin-kiegel: As a sidenote: associated consts would be good to convert [this](https://github.com/colin-kiegel/rust-derive-builder/blob/feature/field-overrides/src/options.rs#L221-L227) `if let {} else if let {}` pattern into a `match` which looks nicer IMO. 

If you prefer a default impl I can change it, but it would still not allow to switch to the nicer match-syntax. :-/

PS: On a high level I am a bit worried about differentiating the attribute names for the struct and field case. On the plus side it's grammatically better, but on the minus side it blocks copy-pasting attributes between levels. I have a slight preference to change that and reduce the complexity of the API. What do you think?

killercup: I think making it less complex is a good idea. We can always overengineer it later 😆

colin-kiegel:
I did some cleanup:
- document setter prefix
- document logging
- simplify API: #[setters] + #[setter] -> #[setter]
- removed unimplemented stubs for getters, implicit options, ...
- enable doc-tests on travis and make them work (requires to disable env_logger)

Note, that my current approach to enable doc-tests is a bit hacky. I introduced a `logging` feature to be able to prevent registering a global `env_logger`. Because in the doc-test context there already seems to be an env_logger registered. It feels like another cargo bug, because I would expect a doc-test to be in some kind of empty sandbox without such collisions. That's why I took a quick route to disable logging without completely disabling it. That is I just don't register another env_logger. But log-calls stay active, since I didn't want to introduce fake log! macros. Maybe I should have called the feature env_logging instead. ..  I hope that makes sense. ^^

https://github.com/colin-kiegel/rust-derive-builder/blob/feature/field-overrides/.travis.yml#L29
https://github.com/colin-kiegel/rust-derive-builder/blob/feature/field-overrides/Cargo.toml#L19-L32
https://github.com/colin-kiegel/rust-derive-builder/blob/feature/field-overrides/src/lib.rs#L216-L230

colin-kiegel:
IMO the next step would be to move to the traditional builder pattern. I haven't started any work on that yet. Do you want to try, or shall I? Shall we merge first and do it in another PR?

killercup:
Re env_logger: How does this work? Cargo itself uses env_logger internally IIRC. Can these conflict or overwrite each other? Or is there a way to call it like "create a new env logger unless one is already running"?

colin-kiegel:
Yes, cargo uses it too. As far as I know proc macros run in a separate
process and can therefore register their own env_logger. Well at least it's
working without trouble.

However in the combination of cargo and doc tests in a proc macro something
seems to go wrong and I find that very surprising and don't quite
understand it. The runtime error message is very clear: It complains about
an env_logger being present already. But I don't know why it only fails in
doc tests but not elsewhere.

My plan is to file this as a potential bug, but I would like to test this
in a reduced toy setup first. I just didn't have the time so far.

Pascal Hertleif <notifications@github.com> schrieb am Sa., 21. Jan. 2017,
15:18:

colin-kiegel:
Yes, cargo uses it too. As far as I know proc macros run in a separate process and can therefore register their own env_logger. Well at least it's working without trouble.

However in the combination of cargo and doc tests in a proc macro something seems to go wrong and I find that very surprising and don't quite understand it. The runtime error message is very clear: It complains about an env_logger being present already. But I don't know why it only fails in doc tests but not elsewhere.

My plan is to file this as a potential bug, but I would like to test this in a reduced toy setup first. I just didn't have the time so far.

colin-kiegel:
PS: Maybe doc-tests reuse previous processes which are therefore 'unclean'?? Just guessing.. ^^

colin-kiegel:
PPS: When I file a ticket I will cc you and a env_logger DEV. Hopefully we will know more then. :-)

colin-kiegel:
ok, got it. 
=> We should add a testcase with more than one builder derive. :-)

In our doc-tests `#[derive(Builder)]` is just called _multiple_ times and only the first doc-test works, all others fail:

```rust
#[proc_macro_derive(Builder, attributes(setter))]
pub fn derive(input: TokenStream) -> TokenStream {
    env_logger::init().unwrap();
   // ...
}
```

If I replace `unwrap()` with `is_ok()` then all doc-tests work fine. I will try to introduce a `is_initialized` boolean and I'll remove the `logging` feature again. I think it's reasonable to always have logging and it only impacts compiler performance.

colin-kiegel:
@killercup the doc-tests issue is fixed and multiple derives are possible now. I needed to make this thread safe, but since I am not an expert here I just used the most restrictive ordering.
https://github.com/colin-kiegel/rust-derive-builder/blob/feature/field-overrides/src/lib.rs#L232-L234

killercup:
@colin-kiegel want to finish this and publish a new version ready for 1.15?

colin-kiegel:
I can get to it tomorrow.

What do you think is more important?
A) get macros 1.1 support out as fast as possible
B) implement the traditional builder pattern first

I am a bit undecided, but with a tendency for B.

killercup:
Depends on how long you plan on needing for B :)

I'd personally aim for getting a new release (with nice release notes) ready for the next This Week In Rust.

PR #35, opened 2017-02-05
Feature/traditional builder

colin-kiegel:
ok, here is a working prototype.

This is what I did:
- changed the attribute namespace from `#[setter]` to `#[builder]`
- changed the pattern option from `#[builder(immutable)]` to `#[builder(pattern="immutable")]`
- changed the pattern option from `#[builder(prefix="with")]` to `#[builder(setter_prefix="with")]`
- introduced customizing the builder name, e.g. `#[builder(name="MyBuilder")]` (defaults to FooBuilder, for a given `struct Foo`)
- the builder struct wraps all field types in options
- the builder type implements a `build` method which returns `Result<Foo, String>`, where the error would be like String::from("some_field must be initialized").

I realised that field overrides don't make much sense for the builder pattern, since it impairs concatenation if e.g. owned and mutable setters are mixed. I think we _could_ restrict the builder pattern option to the struct. What do you think?

For the immutable/mutable pattern I clone each field in the build method. For the immutable pattern I clone the builder in each setter. As of now the builder has an unconditional `#[derive(Clone)]` to support the immutable pattern. In effect this means that all fields need to support clone. For generic structs this also means that any type parameter `T`must have a boundary `where T: std::clone::Clone` on the original struct - since we copy all where clauses from there. This whole thing may be a bit too restrictive and we might want to revisit that later.

I thought about Try/TryFrom
- the user would need to clone before calling TryFrom or Try in the mutable/immutable case (not so good)
-  TryFrom is not yet stable
=> So maybe we just stick with `build()`?

TODO
- I did'nt update the docs beyond simple search-replace
- some refactoring / cleanup here and there

Outlook:
- transfer some whitelisted attributes from `Foo` to `FooBuilder` on both the struct and field level (note: currently we only transfer attributes to the setter methods, like allow camel_case, etc.).
- we could introduce structured errors


colin-kiegel:
What do you think about the API changes?
It's still `#[builder(public)]`  and `#[builder(private)]` right now, despite the fact that there is also `pub(foo)`.

colin-kiegel:
cc @badboy

colin-kiegel:
What I would like to refactor most is [this large function](https://github.com/colin-kiegel/rust-derive-builder/blob/e775e669fb42ff93cb7ea94fb6d822ec933f288a/src/lib.rs#L268-L359). It's basically doing this right now
```rust
fn builder_for_struct(ast: syn::MacroInput) -> quote::Tokens {
    let struct_fields = match ast.body { /* ... */ };

    let mut funcs          = Vec::<quote::Tokens>::with_capacity(fields.len());
    let mut builder_fields = Vec::<quote::Tokens>::with_capacity(fields.len());
    let mut initializers   = Vec::<quote::Tokens>::with_capacity(fields.len());

    for f in struct_fields {
        funcs.push(quote!( /* ... */ ));
        // ...
    }

    quote!( /* ...putting everything together */ )
}
```

**Is there a more idiomatic way to work with `syn`/`quote`?** I wonder if we can implement the `quote::ToTokens` trait for some intermediate representations like `struct BuilderStruct`, `struct BuilderSetterFn` and `struct BuilderBuildFn`. I didn't have time to look at other crates using `syn`/`quote`, but this might be a good idea.

Any ideas and pointing in directions are very welcome. :-)

However I don't think it's a show-stopper right now.

dtolnay:
In serde_derive I have used a style more like:

```rust
let funcs = f_opts(fields, &opts)
    .filter(|&(_. opt)| opt.setter_enabled())
    .map(|(f, opt)| derive_setter(f, opt));
let builder_fields = /* similar */;
let initializers = /* similar */;

quote! {
    /* put it together */
}
```

Why do you need the `build` variable near the bottom? I think it would be clearer in line.

colin-kiegel:
@dtolnay Thx. I used iterators when I only had `funcs`. Now that I also have `builder_fields` and `initializers` I want to share the field options obtained from parsing the field attributes `let f_opts = parse_attributes(&f.attrs)`. That's why I thought it might be better to iterate over the fields only once in a big loop. But I didn't do any benchmark. Alternatively I think I could calculate these field options in advance and store them in a vector. That might even take less memory than storing all those tokens as I do now. :-)

I factored out the build method, because we are already thinking about having different implementations based on additional options.

Thanks you already helped me! :-)

dtolnay:
Try not to worry about benchmarks. I have never seen proc macro expansion significantly add to the compile time - the only thing that matters is the amount of code generated. Focus on making the code easy to understand.

colin-kiegel:
@killercup I will probably merge and publish this evening. ;-)

src/lib.rs
```diff
@@ -19,78 +23,72 @@
...
-//!     }
-//! }
+//! Let's look again at the example above. You can build structs like this:
 //!
-//! fn main() {
-//!     let x = Lorem::default().ipsum("42").build();
-//!     println!("{:?}", x);
-//! }
+//! ```rust,ignore
+//! let x: Lorem = LoremBuilder::default().ipsum("42").build()?
```
killercup: Is is much work to make this a full doc test? I think

```
# #[macro_use] extern crate derive_builder;
# #[derive(Builder)] struct Lorem { ipsum: String }
# fn try_main() -> Result<(), Box<::std:.error::Error>> {
let x: Lorem = LoremBuilder::default().ipsum("42").build()?;
# } fn main() { try_main().unwrap(); }
```

should do it

src/lib.rs
```diff
@@ -19,78 +23,72 @@
...
 //!
-//! # Builder Patterns
-//!
-//! Let's look again at `let x = Lorem::default().ipsum("42").build()`.
-//! Chaining method calls is nice, but what if `ipsum("42")` should only happen if `geek = true`?
+//! Ok, _chaining_ method calls is nice, but what if `ipsum("42")` should only happen if `geek = true`?
 //!
 //! So let's make this call conditional
 //!
 //! ```rust,ignore
```
killercup: Same here, non-ignored doc tests are good doc test :D

src/lib.rs
```diff
@@ -271,96 +280,165 @@ fn filter_attr(attr: &&syn::Attribute) -> bool {
...
-    let name = &ast.ident;
+    let struct_name = &ast.ident;
+    let builder_name = syn::Ident::from(opts.builder_name());
     let (impl_generics, ty_generics, where_clause) = ast.generics.split_for_impl();
-    let funcs = fields.iter().map(|f| {
 
-        trace!("Parsing field {:?}.", f.ident.as_ref().map(|i| i.as_ref()).unwrap_or_default());
+    let mut setter_fns     = Vec::<quote::Tokens>::with_capacity(fields.len());
+    let mut builder_fields = Vec::<quote::Tokens>::with_capacity(fields.len());
+    let mut initializers   = Vec::<quote::Tokens>::with_capacity(fields.len());
```
killercup: Three times? I'd `type TokenVec = Vec::<quote::Tokens>;`this thing 😉

src/lib.rs
```diff
@@ -271,96 +280,165 @@ fn filter_attr(attr: &&syn::Attribute) -> bool {
...
         let f_opts = OptionsBuilder::<FieldMode>::default()
             .parse_attributes(&f.attrs)
-            .with_struct_options(&opts);
-
-        debug!("Setter prefix is {:?}", f_opts.setter_prefix());
+            .build(name, &opts);
 
-        let mut tokens = quote::Tokens::new();
+        trace!("Filtering field attributes");
+        let attrs: Vec<&syn::Attribute> = f.attrs.iter()
```
killercup: …and also `type AttrVec = Vec<&syn::Attribute>;` while you're at it :) (notice the borrow)

src/lib.rs
```diff
@@ -237,9 +246,9 @@ pub fn derive(input: TokenStream) -> TokenStream {
 
     let ast = syn::parse_macro_input(&input).expect("Couldn't parse item");
 
-    let result = builder_for_struct(ast);
+    let result = builder_for_struct(ast).to_string();
 
-    format!("{}", result).parse().expect("Couldn't parse string to tokens")
+    result.parse().expect(&format!("Couldn't parse {:?} to tokens", result))
```
killercup: rustc and cargo usually put code or values in error messages backticks, i.e. ```"Couldn't parse `{:?}` to tokens"```. (Changing here and everywhere else this will require you to adjust a bunch of tests though)

src/lib.rs
```diff
@@ -271,96 +280,165 @@ fn filter_attr(attr: &&syn::Attribute) -> bool {
...
-        if f_opts.setter_enabled() {
-            derive_setter(f, &f_opts).to_tokens(&mut tokens);
-        } else {
-            trace!("Skipping setter.");
-        }
+        setter_fns.push(derive_setter(f, &f_opts, &attrs));
+        builder_fields.push(derive_builder_field(f, &f_opts, &attrs));
+        initializers.push(derive_initializer(f, &f_opts));
+        // NOTE: Looking forward for computation in interpolation
+        // - https://github.com/dtolnay/quote/issues/10
```
killercup: Linking here to trigger Githubs magic cross linkery: https://github.com/dtolnay/quote/issues/10

src/lib.rs
```diff
@@ -271,96 +280,165 @@ fn filter_attr(attr: &&syn::Attribute) -> bool {
...
+    let builder_vis = opts.builder_visibility();
+    let build_fn = {
+        let ref_self = match *opts.field_defaults().setter_pattern() {
+            SetterPattern::Owned => quote!(self),
+            SetterPattern::Mutable => quote!(&self),
+            SetterPattern::Immutable => quote!(&self),
+        };
+        quote!(
+            #builder_vis fn build(#ref_self) -> Result<#struct_name #ty_generics, String> {
+                Ok(#struct_name{
```
killercup: I'd put a space before the opening brace but maybe I'm just a ~~grammar~~ syntax nazi 😆

src/lib.rs
```diff
@@ -271,96 +280,165 @@ fn filter_attr(attr: &&syn::Attribute) -> bool {
...
 
-fn derive_setter(f: &syn::Field, opts: &Options) -> quote::Tokens {
-    trace!("Deriving setter.");
-    let ty = &f.ty;
+fn derive_builder_field(f: &syn::Field, opts: &FieldOptions, attrs: &Vec<&syn::Attribute>) -> quote::Tokens {
+    if opts.setter_enabled() {
+        trace!("Deriving builder field for {}.", opts.field_name());
+        let vis = &f.vis;
+        let ident = &f.ident;
+        let ty = &f.ty;
```
killercup: Hm. Is `let (vis, ident, ty) = (f.vis, f.ident, f.ty)` nicer?

src/lib.rs
```diff
@@ -271,96 +280,165 @@ fn filter_attr(attr: &&syn::Attribute) -> bool {
...
+    trace!("Deriving initializer for {}.", opts.field_name());
+
+    let err_uninitizalied = format!("{} must be initialized", opts.field_name());
     let pattern = opts.setter_pattern();
-    let vis = opts.setter_visibility();
-    let fieldname = f.ident.as_ref().expect(&format!("Missing identifier for field {:?}.", f));
-    let funcname = if opts.setter_prefix().len() > 0 {
-        Cow::Owned(syn::Ident::new(format!("{}_{}", opts.setter_prefix(), fieldname)))
+    let ident = &f.ident;
+    // let attrs = &f.attrs;
```
killercup: Choose one: Commenting out code, or using git 😉

(Or, for consistency, un-comment and write like `let _attrs` to not trigger dead code lint)

src/lib.rs
```diff
@@ -271,96 +280,165 @@ fn filter_attr(attr: &&syn::Attribute) -> bool {
...
+
+    if opts.setter_enabled() {
+        let initializer = match *pattern {
+            SetterPattern::Owned => quote!(
+                    #ident: self.#ident.ok_or(#err_uninitizalied)?,
+                ),
+            SetterPattern::Mutable
+            | SetterPattern::Immutable => quote!(
```
killercup: Do people put the pipe at the beginning of the next or the end of the previous line? (I'd do the latter but have no hard preference.)

src/options.rs
```diff
@@ -15,18 +14,44 @@ impl Default for SetterPattern {
...
+    // e.g. `#[builder(pattern="owned")]` (defaults to mutable)
     setter_pattern: SetterPattern,
-    // e.g. `#[setter(prefix="with")]` (defaults to None)
+    // e.g. `#[builder(setter_prefix="with")]` (defaults to None)
     setter_prefix: String,
-    // e.g. `#[setter(private)]` (defaults to public)
-    setter_public: bool,
+    // e.g. `#[builder(private)]` (defaults to public)
+    setter_vis: syn::Visibility,
+    // the _original_ field name
```
killercup: make all these doc comments, they're great

src/options.rs
```diff
@@ -35,70 +60,101 @@ impl Options {
...
 }
 
-pub trait OptionsBuilderMode {}
+pub trait OptionsBuilderMode {
+    fn parse_builder_name(&mut self, lit: &syn::Lit);
+}
 
 #[derive(Default)]
-pub struct StructMode;
+pub struct StructMode{
```
killercup: put a space before opening brace pls

src/options.rs
```diff
@@ -35,70 +60,101 @@ impl Options {
...
+        self.builder_name = Some(value.clone());
+    }
+}
 
 #[derive(Default)]
 pub struct FieldMode;
 
-impl OptionsBuilderMode for FieldMode {}
+impl OptionsBuilderMode for FieldMode {
+    fn parse_builder_name(&mut self, _: &syn::Lit) {
```
killercup: I tend to name even unused bindings to easer grasp what kind of literal I get here. Maybe `_name`?

src/options.rs
```diff
@@ -35,70 +60,101 @@ impl Options {
...
+pub struct StructMode{
+    builder_name: Option<String>,
+    builder_vis: Option<syn::Visibility>,
+    struct_name: String,
+    struct_vis: Option<syn::Visibility>,
+}
 
-impl OptionsBuilderMode for StructMode {}
+impl OptionsBuilderMode for StructMode {
+    fn parse_builder_name(&mut self, lit: &syn::Lit) {
```
killercup: Why not call `lit` `name`? (It's already of type `Lit`)

derive-builder-test/tests/setter_pattern.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+struct Sit {
+    default: u32,
+}
+
+type ImmutableSetter<T, U> = fn(&T, U) -> T;
+type OwnedSetter<T, U> = fn(T, U) -> T;
+type MutableSetter<T, U> = fn(&mut T, U) -> &mut T;
```
killercup: Nice

README.md
```diff
@@ -5,62 +5,96 @@
...
+                .as_ref()
+                .ok_or(// builder pattern, go, go, go!...
+                       "token must be initialized")?),
+            special_info: Clone::clone(self.special_info
+                .as_ref()
+                .ok_or("special_info must be initialized")?),
+        })
+    }
 }
 ```
```
killercup: Maybe add <https://crates.io/crates/skeptic>?

README.md
```diff
@@ -5,62 +5,96 @@
...
+    pub fn special_info<VALUE: Into<i32>>(&mut self, value: VALUE) -> &mut Self {
+        let mut new = self;
+        new.special_info = Some(value.into());
+        new
+    }
+    fn build(&self) -> Result<Channel, String> {
+        Ok(Channel {
+            token: Clone::clone(self.token
+                .as_ref()
+                .ok_or(// builder pattern, go, go, go!...
```
colin-kiegel: removed - (cargo expand does _weird_ things with comments)

README.md
colin-kiegel: done - skeptic included - great tip!

src/lib.rs
```diff
@@ -19,78 +23,72 @@
 //! # fn main() {}
 //! ```
 //!
-//! `#[derive(Builder)]` will automatically generate a setter method for the `ipsum` field,
-//! looking like this:
+//! ## What you get
 //!
 //! ```rust,ignore
```
colin-kiegel: done - doc test enabled.

but I kept the allow dead_code attr, because I would like to be transparent about the code which is generated.

src/lib.rs
colin-kiegel: done - thx

src/lib.rs
colin-kiegel: done - thx again :-)

src/lib.rs
colin-kiegel: done - I switched from {:?} to `{}` there.

I also changed other places from {:?} to `{:?}`, because most syn-Items only have the fmt::Debug trait.

Until now I prefered {:?} over `{}`, because it is less ambiguous. It's either
```
Couldn't parse "struct \"foo\" {}" to tokens: LexError { _inner: () }
```

or
```
Couldn't parse `struct "foo" {}` to tokens: LexError { _inner: () }
```

The second variant looks nicer - but becomes ambiguous if the token stream itself contains a back tick. => I guess it's a trade off worthwhile.

src/lib.rs
colin-kiegel: done - thx

src/lib.rs
colin-kiegel: done - thx

src/lib.rs
colin-kiegel: done

src/lib.rs
colin-kiegel: done - thx

src/lib.rs
colin-kiegel: done

src/lib.rs
colin-kiegel: I think it's actually more common on the end of the line.

But I thought it would make more sense at the beginning because when you say "A or B" you would put a pause like this "A ... or B" and not "A or ... B". But let's go with the crowd.

src/options.rs
colin-kiegel: done :-)

src/options.rs
colin-kiegel: done

src/options.rs
colin-kiegel: done

src/options.rs
colin-kiegel: done

derive-builder-test/tests/setter_pattern.rs
colin-kiegel: thx. :-)

colin-kiegel:
:tada: Thanks a lot!

Issue #37, opened 2017-02-13
missing_docs


colin-kiegel:
The following fails
```
#![deny(missing_docs)]
#[derive(Builder)]
struct Foo {
    bar: bool
}
```
with `error: missing documentation for a method`.

While doing this we should probably add some other `deny`s to our tests, to be sure.

killercup:
Diesel uses

    #![deny(warnings, missing_debug_implementations, missing_copy_implementations)]

fyi

colin-kiegel:
Thx - is it possible to set that specifically for `cargo test` but not `cargo build` (when applied to the crate itself)?

Use case: Rust might introduce deprecation warnings, but we still want the _build_ to succeed.

killercup:
You mean like `#[cfg_attr(test, allow(missing_docs))]`?

colin-kiegel:
perfect, thanks! :-)

PR #38, opened 2017-02-13
deny missing docs

colin-kiegel:
fix https://github.com/colin-kiegel/rust-derive-builder/issues/37


src/lib.rs
```diff
@@ -362,9 +362,31 @@ fn builder_for_struct(ast: syn::MacroInput) -> quote::Tokens {
...
+    let builder_struct_doc = format!(
+        "Builder for {s}.\n\
+        \n\
+        # Examples\n\
+        \n\
+        ```\n\
+        let builder = <{b} as Default>::default();\n\
+        // .. call some setters on `builder`\n\
+        let x: Result<{s}, String> = builder.build();\n\
+        ```",
```
killercup: I'd write it without the `\n` dance and just live with the shitty indents.

src/lib.rs
```diff
@@ -362,9 +362,31 @@ fn builder_for_struct(ast: syn::MacroInput) -> quote::Tokens {
         )
     };
 
+    let builder_struct_doc = format!(
+        "Builder for {s}.\n\
+        \n\
+        # Examples\n\
+        \n\
+        ```\n\
```
killercup: Add `rust` after the back ticks would make me very happy :)

killercup: Also, can we be sure this will always work as a doc test? (I think so.)

colin-kiegel: Hm isn't `rust` implicit for Examples? http://rustbyexample.com/meta/doc.html
Or do you generally prefer explicitness here?

Surprisingly I can't find this doc test in the [output](https://travis-ci.org/colin-kiegel/rust-derive-builder/jobs/201316301#L347-L464) of `cargo test`. I will double check if the behaviour in a foreign crate is the same and also if MBEs behave differently. We could resort to some trickery, e.g. put the documentation into `examples/builder_struct_doc_template.rs`, import it via `include_str` and find a workaround for `format!` placeholders (`FOO` instead of `{s}`).

killercup: > Or do you generally prefer explicitness here?

Yes.

> Surprisingly I can't find this doc test in the output of cargo test.

I assume you made an example crate for this? Can you find the docs in `cargo doc`?

src/lib.rs
colin-kiegel: :man_dancing: Then I get ..night \n\ ..fever /n/ ..night \n\ ..fev /n/ ...er! \o/

src/lib.rs
colin-kiegel: I can find the documentation in `cargo expand --example missing_docs` ([examples/missing_docs.rs](https://github.com/colin-kiegel/rust-derive-builder/blob/ca762dbfac5733acb6e730e52675d9222f8afc39/examples/missing_docs.rs)). I expected this to behave like a foreign crate, but I can double check. :-)

src/lib.rs
killercup: 🤣 you win

sgrif: You could still put it in a separate file and `include_str!` it

colin-kiegel:
@killercup I added a build step to generate doc tests from the doc templates... _quite an effort_ to test two lines of code ^^
```rust
let builder = <{builder_name} as Default>::default();
// .. call some setters on `builder`
let _x: Result<{struct_name}, String> = builder.build();
```

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Fixed
 
 - use full path for result #39
+- deny missing docs #37
```
killercup: Maybe… "Allow deny missing docs"? 😆

colin-kiegel: Will change that :-)

colin-kiegel:
The thing is, there might be a setter method called `default`. I am not sure if Rust will complain, since `Default::default()` is static.

Do you think `Default::<FooBuilder>::default()` would be better, if it is needed?

colin-kiegel:
Note to myself: Update the doc tpl test case to have a field called `default` :-)

colin-kiegel:
Ok, having a field called `default` indeed leads to this error, which means we might be generating poor documentation for _someone_, sigh.

```
11 |    let builder = FooBuilder::default();
   |                  ^^^^^^^^^^^^^^^^^^^^^ expected 2 parameters
```

I see these options:
1. go with the simple documentation and 
  a. risk documentation error for some people
  b. error on setter name `default`
2. document as `let builder: FooBuilder = Default::default()`
3. document as `let builder = <FooBuilder as Default>::default()`

@killercup My preference is still 3. It just works. No 2 would break as soon as you start chaining setters and a terminal `build()`, 1b is an unnecessary restriction and 1a may lead to confusing error messages if someone tries the example code in this edge case. You are right that No 3 looks overly complicated, but people who don't know this construct might learn something along the way. ;-)

killercup:
True. Let's go with No 3.

PR #40, opened 2017-02-15
Use full path for Clone, Into, Option and Result

### LINKED ISSUE #39: Use full path for Result
I think you should use `std::result::Result` instead `Result` in [this place](https://github.com/colin-kiegel/rust-derive-builder/blob/master/src/lib.rs#L357).
Cause: if declare or `use` custom type `Result` it will throw error: `error[E0244]: wrong number of type arguments: expected 1, found 2`

0g00:
std-only support
Fix #39

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -354,7 +354,7 @@ fn builder_for_struct(ast: syn::MacroInput) -> quote::Tokens {
             SetterPattern::Immutable => quote!(&self),
         };
         quote!(
-            #builder_vis fn build(#ref_self) -> Result<#struct_name #ty_generics, String> {
+            #builder_vis fn build(#ref_self) -> ::std::result::Result<#struct_name #ty_generics, String> {
                 Ok(#struct_name {
                     #(#initializers)*
                 })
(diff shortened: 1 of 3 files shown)


colin-kiegel:
Thx for the PR, looks good.

Do you want to also do full paths for these? (The test case could be called `custom_types` then).
* Option
* String
* Clone
* Into
* default::Default

If not I would accept the PR anyway. :-)

cramertj:
Note that this won't work with `no_std`: for that you'll need `::core::result::Result`.

colin-kiegel:
@cramertj Interesting, would ::core::... always work, or can we cfg_attr around it?

cramertj:
`::core::...` works as long as there's an `extern crate core;` declaration.

0g00:
Do need explicitly add `extern crate core;` into each crate that uses `#[macro_use]
extern crate derive_builder;`?
And what about `String`? without `std` it also requires explicitly add `extern crate collections;`...

colin-kiegel:
@0g00 We don't need to solve both issues with the same PR. You _can_ assume std-only support for the scope of this PR and I will open a separate ticket for no-std support. We might want to update our docs for no-std too (`extern crate core` and `collections`).

colin-kiegel:
Hm, I will try to solve the second issue before a new release. Otherwise we would knowingly introduce a regression since nostd should work now.

Issue #15, opened 2016-08-08
idea: hidden fields


colin-kiegel:
# use case

User wants to expose custom setters instead, which are more complex, do additional checks, etc.
# possible syntax

field annotation:
- hide: `#[builder(hide=true)]`
- unhide `#[builder(hide=false)]`

shorthand
- hide: `#[builder(hide)]`
- unhide: `#[builder]`

# Example

``` rust
#[derive(Builder)]
struct Foo {
    #[builder(hide)]
    foo: Vec<i32>
}
```
# alternatives
- hide
- ignore
- omit
- skip
- no_setter (i.e. `#[builder(no_setter)]` instead of `#[builder(setter(skip))]` ...)
- ...

dpc:
It seems to me lack of this prevents me from using this (BTW. really nice) crate.

The struct that I want to build has way more fields than the ones I want to expose in the builder crate.

colin-kiegel:
@dpc thx for your feedback. Would all hidden fields implement the `Default` trait, or do you want to annotate a field with a custom default specific only to the build method (e.g. `#[builder(hide, default="vec![1,2,3]")])`?

dpc:
The thing is: I'd actually want to annotate/describe the `struct MyStructBuilder` itself, not `MyStruct`. The `struct MyStruct` is fancy, contains data like channels, `Arc`s of weird things, etc. 

What I'd want is:

```
#[derive(Builder)]]
struct MySructBuilder {
   field1 : int, 
   field2 : bool,
   ...
}
```

that would generate a code all the setters and:

```
impl MyStructBuilder {
    fn build(self) -> MyStruct {
          MyStruct::from(self)  // or just self.into() ?
    }
}
```

or something like that.

This way I can describe the builder without impacting how `MyStruct` is supposed to work and look like. Also, if I wanted to I could have multiple different builders for my crate (with some argument support) etc.

colin-kiegel:
@dpc Interesting. The bottomline is that you would write `impl From<MyStructBuilder> for MyStruct` _yourself_, right?

I have the feeling this might actually be a use case for the [accessors](https://github.com/emk/accessors) crate. What if you use that to use `#[derive(setters)]]` instead
```rust
#[derive(setters)]]
struct MySructBuilder {
   field1 : int, 
   field2 : bool,
   ...
}
```

Then impl the `From` trait and treat `.into()` as your build method?

[...]

PR #44, opened 2017-02-19
Feature/skip setter

colin-kiegel:
- skip setters, e.g. `#[builder(setter(skip))]`
  https://github.com/colin-kiegel/rust-derive-builder/issues/15
- deprecated syntax `#[builder(setter_prefix="with")]`,
  please use `#[builder(setter(prefix="with"))]` instead
- Custom derives can't emit nice warnings to the user??
  Hah! Have a look at [this](https://github.com/colin-kiegel/rust-derive-builder/commit/4c712d71b469ff3e544371fc84f5b65320332899) and [this](https://travis-ci.org/colin-kiegel/rust-derive-builder/jobs/203264430#L351-L357)! :-)

### DIFF
--- a/src/deprecation_notes.rs
+++ b/src/deprecation_notes.rs
@@ -1,0 +1,31 @@
+use quote::{Tokens, ToTokens};
+use syn;
+
+/// Deprecation notes we want to emit to the user.
+#[derive(Debug, Default, Clone)]
+pub struct DeprecationNotes(Vec<String>);
+
+impl DeprecationNotes {
+    /// Appends a note to the collection.
+    pub fn push(&mut self, note: String) {
+        self.0.push(note)
+    }
+}
+
+/// This will emit a block which can be inserted into any fn body to emit a
+/// deprecation warning in the downstream crate. Cool stuff. ^^
+/// Proof of concept: https://play.rust-lang.org/?gist=8394141c07d1f6d75d314818389eb4d8&version=stable&backtrace=0
+impl ToTokens for DeprecationNotes {
+    fn to_tokens(&self, tokens: &mut Tokens) {
+        for note in &self.0 {
+            let fn_ident = syn::Ident::new("derive_builder_deprecation_notice");
+            tokens.append(&quote!(
+                {
+                    #[deprecated(note=#note)]
+                    fn #fn_ident() { }
+                    #fn_ident();
+                }
+            ).to_string());
+        }
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -273,6 +273,7 @@ extern crate log;
 extern crate env_logger;
 
 mod options;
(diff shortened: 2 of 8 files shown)


colin-kiegel:
I'm not _really_ sure that I want to deprecate `#[builder(setter_prefix="with")]` in favour of `#[builder(setter(prefix="with"))]`. But it was a good testcase for my idea to model deprecation notes.

Feel free to criticise anything you don't like. My current tendency is to accumulate a few features in the master branch before rolling out the next release. This way we can iterate a little over our API and see how it plays together first. Next in line would be `no_std`, `optional setters` and `CustomDefaults` IMO.

src/lib.rs
```diff
@@ -194,6 +194,34 @@
 //! # fn main() {}
 //! ```
 //!
+//! ## Hidden Fields
+//!
+//! You can hide fields by skipping their setters on the builder struct.
+//!
+//! - Opt-out &mdash; skip setters via `#[builder(setter(skip))]` on individual fields.
+//! - Opt-in &mdash; set `#[builder(setter(skip))]` on the whole struct
```
killercup: Does rustdoc actually render `&mdash;` as HTML entity? Freaky. If not, here, haven an m-dash: `—` 😄

src/deprecation_notes.rs
```diff
@@ -0,0 +1,31 @@
...
+impl DeprecationNotes {
+    /// Appends a note to the collection.
+    pub fn push(&mut self, note: String) {
+        self.0.push(note)
+    }
+}
+
+/// This will emit a block which can be inserted into any fn body to emit a
+/// deprecation warning in the downstream crate. Cool stuff. ^^
+/// Proof of concept: https://play.rust-lang.org/?gist=8394141c07d1f6d75d314818389eb4d8&version=stable&backtrace=0
```
killercup: this is brilliant :O

src/options.rs
```diff
@@ -286,20 +314,129 @@ impl<Mode> OptionsBuilder<Mode> where
...
+            }
+        }
+    }
+
+    /// e.g `setter(skip)` in `#[builder(setter(skip))]`
+    #[allow(non_snake_case)]
+    fn parse_builder_options_list(
+        &mut self,
+        ident: &syn::Ident,
+        nested: &[syn::NestedMetaItem]
```
killercup: FYI, you can use trailing commas in param lists to prevent needless diff churn in the future

src/lib.rs
colin-kiegel: yes, `&mdash;` works in rustdoc - afaik it's a feature of markdown in general. :-)

colin-kiegel:
I have heard of prom-attire but it's currently _not_ documented at all and not even published on crates.io... That's basically all I know right now. :-)

Issue #41, opened 2017-02-16
officially support `#![no_std]`


colin-kiegel:
TODO: Add example + docs for `#![no_std]` scenario.

- We'll need `::core::result::Result` instead of `::std::result::Result` etc.
- ::core::... works iff there's an `extern crate core;` declaration.
- `String` would require to explicitly add `extern crate collections;` (unless we find a way to introduce structured errors #60)

colin-kiegel:
Try not to introduce a regression with #40

colin-kiegel:
published with [v0.4.0](https://github.com/colin-kiegel/rust-derive-builder/releases/tag/v0.4.0)

PR #45, opened 2017-02-20
officially support `#![no_std]` #41

colin-kiegel:
cc @cramertj

This is work in progress and does not work yet.

open questions
- how do we tell cargo or travis that a certain test should only run on nightly rust? (#![feature(collections)] requires nightly)
- something **strange** happens with this code.
```rust
quote!(
    #[cfg(not(no_std))]
    #builder_vis fn build(#ref_self) -> ::std::result::Result<#struct_name #ty_generics, ::std::string::String> {
        /* ... */
    }
    #[cfg(no_std)]
    #builder_vis fn build(#ref_self) -> ::core::result::Result<#struct_name #ty_generics, ::collections::string::String> {
        /* ... */
    }
)
```

The expanded code _only_ contains the first part
```rust
#[cfg(not(no_std))]
#builder_vis fn build(#ref_self) -> ::std::result::Result<#struct_name #ty_generics, ::std::string::String> {
    /* ... */
}
```

It's like the cfg-attributes always evaluate to `#[cfg(not(no_std))]==true` and `#[cfg(no_std)]==false` -- independently of the testcase. I would not be surprised if this was a rust or cargo bug evaluating the cfg-attribute in the wrong context - otherwise I don't understand it.

### DIFF
--- a/derive_builder_core/src/build_method.rs
+++ b/derive_builder_core/src/build_method.rs
@@ -48,6 +48,8 @@ pub struct BuildMethod<'a> {
     pub initializers: Vec<Tokens>,
     /// Doc-comment of the builder struct.
     pub doc_comment: Option<syn::Attribute>,
+    /// Whether the generated code should comply with `#![no_std]`.
+    pub no_std: bool,
 }
 
 impl<'a> ToTokens for BuildMethod<'a> {
@@ -63,12 +65,21 @@ impl<'a> ToTokens for BuildMethod<'a> {
             BuilderPattern::Immutable => quote!(&self),
         };
         let doc_comment = &self.doc_comment;
+
+        let (result, string) = if self.no_std {(
+            quote!(::core::result::Result),
+            quote!(::collections::string::String),
+        )} else {(
+            quote!(::std::result::Result),
+            quote!(::std::string::String),
+        )};
+
         if self.enabled {
             trace!("Deriving build method `{}`.", self.ident.as_ref());
             tokens.append(quote!(
                 #doc_comment
                 #vis fn #ident(#self_param)
-                    -> ::std::result::Result<#target_ty #target_ty_generics, ::std::string::String>
+                    -> #result<#target_ty #target_ty_generics, #string>
                 {
                     Ok(#target_ty {
                         #(#initializers)*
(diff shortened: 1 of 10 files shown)


colin-kiegel:
A workaround would be having an explicit `#[builder(no_std)]` attribute - but I feel like we should try to get the conditional compilation approach working.

Suggestions are welcome - I am kind of stuck right now. :-)

This is what I mean: [`cargo expand --test no_std`](https://gist.github.com/colin-kiegel/c451cad4f544cde9e33b9034766c70ce#file-no_std_expanded-rs-L40-L47)

colin-kiegel:
Ok, I have now worked around the strange issues with emitting `#[cfg(this)] ... #[cfg(that)] ...` code.
The new approach requires to annotate the struct with `#[builder(no_std)]`.

colin-kiegel:
Oh, interesting - compiletest seems to [not support `#[no_std]`](https://travis-ci.org/colin-kiegel/rust-derive-builder/jobs/215042549#L553-L555) that easily... Let's see.

derive_builder_core/src/build_method.rs
```diff
@@ -63,12 +65,21 @@ impl<'a> ToTokens for BuildMethod<'a> {
...
+            quote!(::std::string::String),
+        )};
+
         if self.enabled {
             trace!("Deriving build method `{}`.", self.ident.as_ref());
             tokens.append(quote!(
                 #doc_comment
                 #vis fn #ident(#self_param)
-                    -> ::std::result::Result<#target_ty #target_ty_generics, ::std::string::String>
+                    -> #result<#target_ty #target_ty_generics, #string>
```
killercup: This is the reason `extern crate collections` is required, right? Can we work around that? E.g., not give string error messages when using `no_std` but more simple ones?

colin-kiegel: Yep, I was thinking about that too.

I would say, that we could implement that as a separate feature. In fact this could just mean to provide more structured errors. I think I'm going to open an issue for that and link it here. How does that sound? :-)

killercup: 👍

PR #48, opened 2017-03-11
refactor code generation

colin-kiegel:
cc @killercup @dtolnay _if_ you are interested in this:

I re-factored the code generation of `derive_builder` and I am looking for
some general feedback. :-)

Since I don't have _that_ much programming experience, I may have made a
questionable design choice here or there. Hence I am primarily interested
in whether the general architecture looks ok-<i>ish</i> to other people on a
high-level.

The interesting bits are in `src/lib` and `derive_builder_core/src`.

(I also welcome nit-picking, if anyone wants to take that extra time. :smile:)

# Motivation and Goals

There are _a lot_ of feature ideas for this crate (like a typesafe builder
pattern, custom defaults, and a lot of smaller features).
This will (most likely) lead to a huge combinatorial complexity of options.

The current design would probably get very messy over time. It also lacks unit
tests.

# Approach

In order to prepare this crate for future features, I tried to *modularize* the
code generation and allow for better unit testing.

- I defined helper structs, like `Builder`, `BuilderField`, `Setter`,
  `BuildMethod`, `Initializer`, which implement `quote::ToTokens`.
- These helper structs each come with their own unit tests.
- I moved all the code generation into a helper crate `derive_builder_core`,
  which is not a proc-macro, but a traditional library.
  - This helped to establish a clear separation.
  - New features can be tested and iterated in the core crate for a while
    before implementing a public API in the main crate.
  - It would allow the main crate `derive_builder` to consume the core lib
    in a _build step_ to implement the builder pattern for parsing options
    (..maybe).
  - Lastly it could help to break cyclic dependencies in the future, if
    `derive_builder` wants to depend on a crate, which in turn wants to depend
    on `derive_builder`.
- I added compile-tests - *Yippie*.
- No new features (in this PR).

# Exclusion

The focus of this re-factoring was _exclusively_ on the code generation side.
I moved some bits of the options parsing around, but that didn't get much love.
My plans are to revisit that with another re-factoring in the future. Ideas here
are to take a look at [`prom_attire_rs`] (which is not on crates.io
unfortunately), or [`synom`] and maybe introduce a code-gen build step based on
`derive_builder_core`.

=> IMO you _can exclude_ the whole `src/options` module for the purpose of this
review. Of course we can already discuss about the future direction here, but I
would probably like to _implement_ that later. :-)

# Alternatives

- Leave it as one crate, since it is uncertain if there really will be a build
  step using `derive_builder_core` in the future. There has also not been
  a problem with cyclic dependencies so far.

[`prom_attire_rs`]: https://github.com/Nemo157/prom-attire-rs/
[`synom`]: https://docs.rs/synom


colin-kiegel:
PS: The new docs are temporarily here http://remote.kiegel.eu/derive_builder/derive_builder_core/index.html

killercup:
Woah, this is huge. I'll try to have a look at it this evening.

From the first few lines I read: You might be able to add `#[deny(warnings, missing_docs)]` to the main and core crate with little effort. Also, the convention in `std` is to always write `# Examples` (plural).

colin-kiegel:
@killercup I'm thinking about your suggestion to move derive_builder into a sub-directory.
- I think I would like to keep the CHANGELOG.md visible at the top-level.
- I'd probably drop the changelog of the core crate then. I was not even sure, if this is worth the effort and I would be inconsistent to have one changelog within and the other one outside of the crate.
- What about the licenses? Should they be duplicated on the top level?

In the meantime I'll give it a shot and test how a virtual Cargo manifest works with my atom IDE.

colin-kiegel:
Ok.. git seems to support symlinks. Hooray. :-)

killercup:
Yep, symlinks for licenses is a good idea. I think Diesel just has all
these meta files in the root and references those from the individual
Cargo.tomls.

I recently saw `cargo build --all`, btw. If it doesn't work well with Atom,
give vscode a try ;) (workspace stuff is not well supported there as well,
but it's a really great editor :D)

Colin Kiegel <notifications@github.com> schrieb am So. 12. März 2017 um
22:34:

colin-kiegel:
Ok, after four attempts I got travis working with the virtual workspace (except for the nightly bug). LOL.

Are we ready to merge then? :-)

colin-kiegel:
closing in favour of #50 :-)

PR #49, opened 2017-03-12
Things I noticed while reviewing

killercup:
⚠️ Targets #48's branch! Feel free to squash.


killercup:
If nightly cargo was working, I'd've run clippy as well…

killercup:
> If I understand the impact correctly, this would also fail a build, where derive_builder is only a dependency. 

Not if I understand [RFC 1193] correctly: Cargo automatically passes `--cap-lints allow`, check `cargo clean; cargo build --verbose`.

[RFC 1193]: https://github.com/rust-lang/rfcs/blob/d8d2de4aa9648dfd6f7bcaee023db9fd7db2da95/text/1193-cap-lints.md

> Both should fail travis builds

You mean our travis builds? We'd need to have a compile step for each sub crate, then (because of cap-lints). Nice CI script, btw, using case is a great idea!

Speaking of CI, the nightly stuff is currently broken because of <https://github.com/rust-lang/cargo/issues/3819>.

PR #50, opened 2017-03-13
Feature/derive builder core rebased

colin-kiegel:
cc @killercup 

I made so unbelievably many iterations on `.travis.yml` in the previous PR #48, that I decided to rebase it. ^^ This branch is now identical with the final state of the previous PR.

I ended up with a helper script [`.travis-run-all.sh`](https://github.com/colin-kiegel/rust-derive-builder/blob/feature/derive_builder_core_rebased/.travis-run-all.sh), which produces [nice output](https://travis-ci.org/colin-kiegel/rust-derive-builder/jobs/210585145#L372-L381) and accumulates the exit code. :tada:

This can be used _like this_ in [`.travis.yml`](https://github.com/colin-kiegel/rust-derive-builder/blob/feature/derive_builder_core_rebased/.travis.yml#L45-L72)
```bash
case "$JOB" in
  # ..
  style_check)
    commands=(
      "cd derive_builder_core && cargo clippy -- -Dclippy"
      "cd derive_builder      && cargo clippy -- -Dclippy"
      "cd derive_builder_test && cargo clippy -- -Dclippy"
      "cd derive_builder_core && cargo fmt -- --write-mode diff"
      "cd derive_builder      && cargo fmt -- --write-mode diff"
      "cd derive_builder_test && cargo fmt -- --write-mode diff"
    )
    ;;
  #..
esac
./.travis-run-all.sh "${commands[@]}"
```

Note that `rustfmt` suggests [unsound changes](https://travis-ci.org/colin-kiegel/rust-derive-builder/jobs/210585145#L483-L484) that would break the build. And some other changes that [I don't like](https://travis-ci.org/colin-kiegel/rust-derive-builder/jobs/210585145#L463-L465). IMO it'd be ok to deviate from rustfmt in some edge cases - as long as they don't get too many.


killercup:
You could've just forced pushed to #48 :)

That travis.yml looks pretty cool :) My most recent idea on how to make such script fancy was [this](https://github.com/diesel-rs/diesel/pull/556), by the way. Might be overkill for this, though.

I wouldn't run rustfmt… but mostly because I don't like the current style that much ;) Did you try adding a config?

You want to merge this right now? :)

colin-kiegel:
merged! :tada: :-)

Btw. my plans are to concentrate on this batch of features in the next days/weeks, like
- no_std
- custom_defaults and optional setters
- require to opt-into setter type conversion (breaking change - but I think `std` does not have generic functions `where T: Into<Foo>` a lot, IMO it's therefore more idiomatic to have it _off_ by default)
- options for the build fn: skip/rename/visibility, to allow for more flexibility to create a custom build fn and potentially wrapping a hidden auto-generated one (which should then be private)

Roughly until the end of march I would like to release a new version with all of this. So if you see a PR in the meantime it's not very urgent. ;-)

PR #51, opened 2017-03-19
put #[deny(warnings)] behind a feature flag

colin-kiegel:
- restore default warning behaviour for cargo,
  which is useful for local development
- travis builds will still deny warnings,
  but only on nightly, which seems sufficient.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ rust:
 matrix:
   include:
     - rust: nightly
-      env: JOB=build CARGO_FEATURES=dev_nightly
+      env: JOB=build CARGO_FEATURES="dev_nightly deny_warnings"
     - rust: nightly
       env: JOB=style_check CARGO_FEATURES=dev_nightly
   allow_failures:
--- a/derive_builder/Cargo.toml
+++ b/derive_builder/Cargo.toml
@@ -23,6 +23,7 @@ proc-macro = true
 
 [features]
 dev_nightly = ["compiletest_rs"]
+deny_warnings = []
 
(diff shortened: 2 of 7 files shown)


killercup:
How many warning are you getting? IMHO `deny(warnings, missing_docs)` is a pretty good default to have _in a library_. For a macro crate it might not matter as much.

colin-kiegel:
This PR is supposed to prevent me from accidentally committing `// #[deny(warnings)]`. :-)

My development workflow is that I often want to quickly run tests on code which has still warnings (dead code, etc.), which I will definitely fix - but later. I only need this when I am in the middle of something.

I noticed that I started temporarily commenting out `deny(warnings)`, which would not be neccessary with a feature-flag. I would probably set my git-push hook to run cargo clean && cargo test with that feature-flag, too.

Don't you ever comment out `deny(warnings)`?

killercup:
Hm, I don't get that many rustc warnings usually (i.e. without clippy), dead code is probably the most common one. But I agree, when writing lots of new code, I usually comment out the denys, especially as missing_docs is not something I want to fix before I'm sure I'm keeping the items. So far I've managed to always uncomment it, though. And when making changes it's nice to have the compiler be as strict as possible.

colin-kiegel:
I have a keyboard shortcut to jump from warning to warning in my IDE and so far I always fixed every warning without `deny(warnings)`, so having it in travis (and my git push hook) would be sufficient for me. Everything else doesn't help me personally. It might only help other future contributors, who might be surprised that travis fails on warnings.

Issue #36, opened 2017-02-12
optional setters


colin-kiegel:
Right now all setters are mandatory. I.e. the build method of `derive_builder` will error if not all fields are initialized.

# use case

User wants to make some setters optional. If no value is provided `Default::default()` should be used instead.

Outlook: [custom defaults](https://github.com/colin-kiegel/rust-derive-builder/issues/3)

# possible api

field annotation:
- optional: `#[builder(optional="true")]` or `#[builder(mandatory="false")]` 
- mandatory `#[builder(optional="false")]` or `#[builder(mandatory="true")]`

shorthand
- optional: `#[builder(optional)]`
- mandatory: `#[builder(mandatory)]`

# example

``` rust
#[derive(Builder)]
struct Foo {
    #[builder(optional)]
    foo: bool
}

#[test]
fn test() {
    let x = FooBuilder::default().build().unwrap();
                                       // ^~~ currently panics with `foo uninitialized`
    assert_eq!(x.foo, false);
}
```

# open questions

- API (see discussion)
- What should be the default behaviour? (See [this suggestion](https://github.com/colin-kiegel/rust-derive-builder/issues/36#issuecomment-281347160) to switch back to "optional by default").

# discussion

Conceptually the hardest part is to pick a good api (=annotation). :-)

This also has significant overlap with the idea to have [custom defaults](https://github.com/colin-kiegel/rust-derive-builder/issues/3), e.g. I would expect that `custom defaults` implicitly make a setter optional. `custom defaults` wouldn't even make sense for mandatory setters. You could argue that an explicit notation for `optional fields` is not needed.

However most types implement `Default`, so a custom default may be unnecessary. Additionally the `optional field` annotation could be used for the whole struct, instead of individual fields - thereby inverting the opt-in to opt-out for that struct. In these cases an optional field annotation is more convenient than custom defaults - but it doesn't remove the overlap.

Now comes the hard part.

Should it be ...
- `optional="true"` vs. `optional="false"`
- .. the above with mandatory instead of optional
- `optional` vs `mandatory`?
- or the whole lot: `optional`, `optional="true"`, `mandatory="false"` vs `mandatory`, `mandatory="true"`, `optional="false"`
- or something more _extensible_ like `style=optional` vs. `style=mandatory`  (where `style` is only a placeholder name)?

Right now I can't think of a reason to extend the binary set `optional` and `mandatory`. 

[...]

PR #53, opened 2017-03-20
implement custom defaults #3

colin-kiegel:
Current implementation support:
- `#[builder(default)]` (without value) to delegate to `Default` 
- or `#[builder(default="42")]` where `42` could be replaced with arbitrarily complex Rust code, as long as it is properly escaped. However it would be recommended to delegate to helper functions for anything fancy.
- This annotation can be used on the struct level, too. Overrides on the field level are still possible.

# Example: Delegated Default

```rust
#[derive(Builder)]
struct Lorem {
    ipsum: String,
    // Custom defaults can delegate to helper methods
    // and pass errors to the enclosing `build()` method via `?`.
    #[builder(default="self.default_dolor()?")]
    dolor: String,
}

impl LoremBuilder {
    // Private helper method with access to the builder struct.
    fn default_dolor(&self) -> Result<String, String> {
        match self.ipsum {
            Some(ref x) if x.chars().count() > 3 => Ok(format!("dolor {}", x)),
            _ => Err("ipsum must at least 3 chars to build dolor".to_string()),
        }
    }
}
```

# Comparison with Serde

Comparison: [This is](https://serde.rs/attr-default.html) how serde does it
* `#[serde(default)]` delegates to `Default`
* `#[serde(default = "some::Function::foo")]` delegates to a function `foo` via UFCS
* it looks like serde does not support arbitrary expressions, but enforces delegation instead
* it looks like arguments on the functions are not supported as well

I feel like the semantics of `#[serde(default = "some::Function::foo")]` would be too restrictive for derive_builder.

* I suspect a lot of _simple_ cases where delegation to `some::Function::foo` would be overkill and `default="42"` is really all you want
* For _complex_ cases it would seem beneficial to delegate to _methods_ on `FooBuilder` which can inspect all builder fields.

It would have been nice to share the semantics with a popular crate like serde. But this would have significant drawbacks IMO.

# Questions

1. My main design question is, whether we should try to enforce delegation for complex expressions or make it more comfortable somehow - or leave it as is (i.e. up to the author).
-> We could try to parse the expression with `synom` and whitelist only a small subset of operations.
-> There could perhaps be a second attribute `default_fn="foo"` which would correspond to the current `default="self.foo()?"` and mimic serde a bit. But this wouldn't safe many characters, but looks like more to explain.
-> Related to 'enforcing things', I experimented a bit with wrapping the default expression either in a block `{ ... }` or closure `|| { ... }`. The closure is the most restrictive option for sandboxing I could think of, but it still allows at least read access to the environment. The current implementation uses the block, which is most permissive w.r.t. manipulating the environment (including moving values out of self and manipulating control flow, e.g. via `?`). 

2. Right now `#[builder(default="")]` does _not_ unset to "no default", but is equivalent to `#[builder(default)]`, i.e. delegate to `Default`. I think I'll add a special case to unset here.

3. `derive_builder_core` has [unused code](https://github.com/colin-kiegel/rust-derive-builder/blob/ef1dd5187cf5dc222c7a0921f6e1d62cac2822e6/derive_builder_core/src/block.rs#L5-L77) from the experiments to wrap things in a closure. Should I remove it? Or is it ok, to stay around for a while?


colin-kiegel:
cc @killercup how do you think about the questions above?

colin-kiegel:
Oh, and there is the overlap with the idea of making fields `optional`. The build method currently fails on uninitialized fields, unless a default was specified. So far so good and this is 95% of [optional setters #36](https://github.com/colin-kiegel/rust-derive-builder/issues/36).

But if all setters are optional - what about the return type of `build`? Should it switch from `Result<T, _>` to `T` implicitly?

After thinking it through quite a bit, I'd say: The return type of `build` should be specified _explicitly_ e.g. via `#[builder(infallible)]` - modulo bikeshedding about the name. Explicitness would also come in handy if we introduce some hooks for custom validation of input `x >= 0 && x < 100` and the like.

So my current thinking is to close #36 with this PR and create a new ticket `infallible build method` instead, where this discussion can be continued.

killercup:
I'll try to review this later today or tomorrow afternoon

derive_builder/src/options/field_options.rs
```diff
@@ -36,12 +38,21 @@ impl FieldOptions {
         }
     }
 
+    /// Panics
```
killercup: Add a description first – "Panics" is not a very _calming_ summary when you see this method in Rustdoc! :) Maybe "Represent this as an `Initializer`". (And then you can also make "Panics" a level 1 heading ~~to give it the font size it deserves~~)

Oh, and I would use `as_initializer` as you don't consume self but return a struct that references self.

derive_builder/tests/custom_default.rs
```diff
@@ -0,0 +1,110 @@
...
+        escaped_default: String,
+        #[builder(default=r#"format!("Hello {}!", "World")"#)]
+        raw_default: String,
+        #[builder(default=r#"format!("{}-{}-{}-{}",
+                             Clone::clone(self.required
+                                .as_ref()
+                                .ok_or("required must be initialized")?),
+                             match self.explicit_default { Some(ref x) => x, None => "EMPTY" },
+                             self.escaped_default.as_ref().map(|x| x.as_ref()).unwrap_or("EMPTY"),
+                             if let Some(ref x) = self.raw_default { x } else { "EMPTY" })"#)]
```
killercup: TIL you can use raw strings in attributes

derive_builder_core/src/block.rs
```diff
@@ -0,0 +1,196 @@
...
+        tokens.append(quote!(
+            { #( #inner )* }
+        ));
+    }
+}
+
+/// Errors
+///
+/// When `expr` cannot be parsed as `Vec<syn::TokenTree>`, e.g. unbalanced
+/// opening/closing delimiters like `{`, `(` and `[` will be _rejected_ as parsing error.
```
killercup: would put this on the from_str method itself

colin-kiegel:
@killercup thx for your review

- I decided to explicitly deny `#[builder(default="")]` for now (this way we can decide on a meaning later). Note that both `#[builder(default=None)]` and `#[builder(default=false)]` won't work, so there will currently not be a way  to opt-out of inherited defaults. But that seems ok for the moment.
- implemented your other suggestions

Will merge if travis is satisfied too. :-)

PR #57, opened 2017-03-25
better diagnostics on panics 

colin-kiegel:
- BEFORE: Unknown setter option `xkrz`.
- AFTER: Unknown setter option `xkrz` on field `foo`.

### DIFF
--- a/derive_builder/src/options/mod.rs
+++ b/derive_builder/src/options/mod.rs
@@ -156,7 +156,7 @@ impl<Mode> OptionsBuilder<Mode> where
 
     /// e.g `private` in `#[builder(private)]`
     fn parse_builder_options_word(&mut self, ident: &syn::Ident) {
-        trace!("Parsing word `{:?}`", ident);
+        trace!("Parsing word `{}`", ident.as_ref());
         match ident.as_ref() {
             "public" => {
                 self.setter_public(true)
@@ -171,7 +171,7 @@ impl<Mode> OptionsBuilder<Mode> where
                 self.default_expression(DefaultExpression::Trait)
             },
             _ => {
-                panic!("Unknown option `{:?}`", ident)
+                panic!("Unknown option `{}`", ident.as_ref())
             }
         };
     }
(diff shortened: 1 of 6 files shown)


derive_builder/src/options/mod.rs
```diff
@@ -233,14 +274,12 @@ impl<Mode> OptionsBuilder<Mode> where
...
-                    // setters implicitly enabled
-                    if self.setter_enabled.is_none() {
-                        self.setter_enabled(true);
-                    }
                 },
-                syn::NestedMetaItem::Literal(ref _lit) => {
-                    // setters explicitly enabled
-                    self.setter_enabled(true);
+                syn::NestedMetaItem::Literal(ref lit) => {
+                    error!("Expected NestedMetaItem::MetaItem, found `{:?}`.", x);
```
killercup: Is this user-visible by default? (I don't know what log-level is the default.) If so, I'd try to explain how that happens and how to fix it, not throw a "NestedMetaItem::MetaItem" in their face ;)

colin-kiegel: No `error!` is not user visible, like `trace!` and `warn!` etc. You have to `RUST_LOG=derive_builder=trace cargo test` in order to see this on `stdout`. I consider it debug output. :-)

Btw. just in this moment I am writing the next pull request, which would put all this debug output behind a `logging` feature flag. Since we have unit tests for code generation I found that the logging became far less important. I may even remove it completely one day.

colin-kiegel:
The missing test is a good hint, but it seems that [`compiletest-rs`](https://crates.io/crates/compiletest_rs) doesn't handle plugin panics very well. I only get something generic like "proc-macro derive panicked", like [here](https://github.com/colin-kiegel/rust-derive-builder/blob/aa49820fff7996f3df449318a8ad11037a783acd/derive_builder_test/tests/compile-fail/deny_empty_default.rs#L7).

So it seems like your ok with this PR?

I'll merge then. :-)

colin-kiegel:
:tada:

PR #58, opened 2017-03-25
opt into generic setters

colin-kiegel:
This seems more idiomatic to me:
- setter conversions are now off by default, you can opt-into via `#[builder(setter(into))]`

=> breaking change — indeed I had to adapt a lot of testcases, which implicitly relied on generic setters.

---
PS: based on pr #57, so only the last commit is relevant here.

### DIFF
--- a/derive_builder/CHANGELOG.md
+++ b/derive_builder/CHANGELOG.md
@@ -6,7 +6,7 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - skip setters, e.g. `#[builder(setter(skip))]`
-- custom defaults, e.g. `#[builder(default="42")]`
+- default values, e.g. `#[builder(default="42")]` or just `#[builder(default)]`
 
 ### Changed
 - deprecated syntax `#[builder(setter_prefix="with")]`,
--- a/derive_builder/CHANGELOG.md
+++ b/derive_builder/CHANGELOG.md
@@ -11,6 +11,8 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 ### Changed
 - deprecated syntax `#[builder(setter_prefix="with")]`,
   please use `#[builder(setter(prefix="with"))]` instead
+- setter conversions are now off by default, you can opt-into via
+  `#[builder(setter(into))]`
 
 ### Fixed
 - use full path for result #39
--- a/derive_builder/README.md
+++ b/derive_builder/README.md
@@ -17,6 +17,7 @@
 extern crate derive_builder;
 
 #[derive(Default, Builder, Debug)]
+#[builder(setter(into))]
 struct Channel {
     token: i32,
     special_info: i32,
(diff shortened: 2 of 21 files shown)


colin-kiegel:
As mentioned before this is one of the things I wanted to do before a new release, to get this breaking change behind us. :-)

killercup:
Can you describe why you consider this more idiomatic? Because the conversions are implicit?

Maybe I'm not the most objective person to comment on this… I really like to use the (infallible) conversion traits _everywhere_, as they allow you (our users) to write much more concise code :)

Re breaking change: This would not be breaking if you turn this into an opt-out.

colin-kiegel:
Ok, this is why I _think_, it's not idiomatic to have setters generic over `Into`

Generic setters can lead to surprises and inference problems. Also I'm having a hard time to find this pattern in `std`.

- Surprise: You write `builder.height_in_meters(height)`. Looks good and compiles. Now `height` might be a new type `Inch(u64)` which implements `Into<u64>` for convenience. But `height.into()` would correspond to _height in inches_, which is not what you are supposed to put into the setter. So setter type conversions could weaken the new-type pattern IMO.
- Inference: For two types `A` and `B` there can only be one `impl Into<B> for A`. So for known types `A` and `B`, Rust can find the right implementation. But e.g. `builder.foo(bar.collect())` would have a problem, where `bar` is some Iterator. Now you can of course annotate `collect`, but what happens if `builder.foo(bar.baz())`, where `fn baz<T: Baz>(&self) -> T`. Now if we only have one type `B` that implements both `Baz` and `Into<Foo>`, then inference should also be possible. But this might break at any time when another Type `B2` starts implementing `Baz` as well. I think this is what is meant in https://github.com/rust-lang/rust/issues/36443.
- Last but not least, most APIs in `std` [seem to be not generic over Into](https://doc.rust-lang.org/std/time/struct.Duration.html). The most prominent 'counter-example' would be [Iterator methods](https://doc.rust-lang.org/std/path/struct.Iter.html), where 11 are generic over `IntoIterator`. But `IntoIterator` is different than `Into<T>`, because it uses associated types instead of a type parameter. I.e. it would probably have less inference problems, I guess.

With default settings, I would like `derive_builder` to generate code which looks like `std`.

colin-kiegel:
PS: all of that does _not_ mean that I don't find `Into` conversions useful here or there.

killercup:
Those are great points! Especially preventing the inference our users expect can do more harm than good. I don't want to stand in the way of this. The most visible difference is that users switching to a new version (and don't opt-in) need to paste in `.into()` a bunch of times. That's okay.

The "Surprise" example took me by surprise (:wink:), by the way. This is the exact use case I'd think a conversion trait is really useful: Let's imagine you change your struct from a "dumb" `u64` to a `Meter` (from something like [dimensioned]). To "strengthen the new-type pattern" so to speak (:wink:²). Then passing in an `Inch(14.2)` would work as expected without needing to manually call `.into()`. But again, it's just an `.into()`. Also interesting to see you link to `std::path` as an example, since `std::path::Path` is one of those types I've only ever used as `P: AsRef<Path>` (like [`File::open`]). 😄

[dimensioned]: https://github.com/paholg/dimensioned
[`File::open`]: https://doc.rust-lang.org/std/fs/struct.File.html#method.open

colin-kiegel:
:tada:

PR #59, opened 2017-03-25
feature flag `logging`

colin-kiegel:
- put logging behind a feature flag to improve compile time,
  since we have unit tests, this is far less important than before.

**Open Question**: I wonder how I should adapt the [travis configuration](https://github.com/colin-kiegel/rust-derive-builder/blob/feature/logging_optional/.travis.yml#L53-L55) to test for features.
- Would it be sufficient to just test `--features "compiletests logging"` on nightly?
- Or should I have one run with default features, and one with full features (except dev_nightly) on each rust release?

### DIFF
--- a/derive_builder_core/src/log_disabled.rs
+++ b/derive_builder_core/src/log_disabled.rs
@@ -6,28 +6,28 @@ macro_rules! log_enabled {
     ($( $x:tt )*) => { false }
 }
 
-// delegate to format and throw away the result to avoid `unused variable` lints.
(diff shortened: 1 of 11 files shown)


derive_builder_core/src/log_disabled.rs
```diff
@@ -0,0 +1,33 @@
+/// Overrides for https://docs.rs/log/#macros
```
killercup: Heh, that's clever!

Issue #61, opened 2017-04-05
#[builder(default)] on struct-level does not use Default impl of the struct


TedDriggs:
rustc version: 1.17.0-nightly (6eb9960d3 2017-03-19)
derive_builder version: 0.4.0

When a struct has a non-derived `impl Default`, those values do not get used in the output of the builder.

**Repro**
```rust
#[macro_use]
extern crate derive_builder;

#![cfg(test)]

#[derive(Debug, Clone, PartialEq, Eq, Builder)]
#[builder(default, setter(into))]
pub struct Example {
    pub a_flag: bool,
    pub limit: Option<String>,
    pub third: String,
}

impl Default for Example {
    fn default() -> Self {
        Example {
            a_flag: false,
            limit: Some("20".into()),
            third: "Hey".to_string(),
        }
    }
}

fn main() {
    let struct_update = Example {
        a_flag: true,
        ..Example::default()
    };
    
    let builder = ExampleBuilder::default().a_flag(true).build().unwrap();
    
    assert_eq!(builder, struct_update);
}
```

This panics with the following error:

```
panicked at 'assertion failed: `(left == right)` (left: `Example { a_flag: true, limit: None, third: "" }`, right: `Example { a_flag: true, limit: Some("20"), third: "Hey" }`)'
```

**Expected:** The resulting structs are equal.

colin-kiegel:
This is an interesting proposal, but it would be a breaking change.

Right now `#[builder(default)]` on the _struct_ level has a very different meaning - it is just sugar for having `#[builder(default)]` _on each field_.
```rust
#[derive(Builder)]
pub struct Example {
    #[builder(default)]
    pub a_flag: bool,
    #[builder(default)]
    pub limit: Option<String>,
    #[builder(default)]
    pub third: String,
}
```
I think the current meaning can also be useful in some cases. I'm not sure which one is more important and I would like to hear more opinions about it.

cc @killercup do you have an opinion about this?

TedDriggs:
I'd interpreted [this comment](issue) on #3 to mean emulating serde's behavior, which does use the `Default` implementation. IMO the by-field approach is the far less common one, and that should be explicitly opted into either by adding that annotation to each field or using something like `#[builder(default(by_field))]`.

[issue]: https://github.com/colin-kiegel/rust-derive-builder/issues/3#issuecomment-284900599

TedDriggs:
[...]

PR #62, opened 2017-04-06
#[builder(default)] now honors struct impls

TedDriggs:
Addresses issue #61.


colin-kiegel:
Cool, thanks again you for your initiative. :-)

Here are some thoughts and wishes for this PR:

* As mentioned in [this comment](https://github.com/colin-kiegel/rust-derive-builder/issues/61#issuecomment-292281744), please put the new behaviour behind a opt-in feature-flag, e.g. `struct_default`?
* It should then be sufficient to have two enum-variants for `DefaultExpression`.
* You can emit a deprecation note [like this](https://github.com/colin-kiegel/rust-derive-builder/blob/df615d26d4e0015acade5f6584c2da65d41a3976/derive_builder/src/options/mod.rs#L207-L210). :-)
  I'm thinking about a message roughly like this
```
warning: the meaning of `#[builder(default)]` on the struct level will change in the
next version (see https://github.com/colin-kiegel/rust-derive-builder/issues/61 for
more details). To squelch this message and adopt the new behavior now, compile
`derive_builder` with `--features "struct_default"`.
```

* I might want to move the field `default_expression` [from `OptionsBuilder` to `OptionsBuilderMode`](https://github.com/colin-kiegel/rust-derive-builder/blob/df615d26d4e0015acade5f6584c2da65d41a3976/derive_builder/src/options/mod.rs#L33-L52
). The first is dedicated to all shared settings, which are also inherited from struct-level to field-level. The latter captures all the things different for struct/field level. The `default` attribute will now be in-between somehow. On the one hand it is shared, but on the other hand it is not inherited anymore and has a completely different implementation on the code generation side. I have a tendency to move this field to `OptionsBuilderMode` right now, but I may have to see the final code here.
* Oh, and I would also appreciate an entry in the Changelog and docs.

Let me know if you need any help/advice.

colin-kiegel:
one addition: If it's not asking too much I would prefer if the generated code was something like
```rust
Foo {
    foo: 42,
    ..::std::default::Default::default()
}
```

instead of 
```rust
let __default = ..::std::default::Default::default();
Foo {
    foo: 42,
    bar: __default.bar,
}
```

TedDriggs:
I actually tried to do that, but I couldn't figure out a way for the macro at compile time to figure out which fields would be populated at runtime. I'm almost done with the other changes; will have a new version up soon.

TedDriggs:
I've been testing this more locally, and found that I only see the deprecation warning when I'm working within `rust-derive-builder`; a local crate that references `derive_builder` by path compiles with no messages. I've checked `cargo expand` and don't see the `#[deprecated(...)]` annotation, but that's also true when I compile the crate's tests and the message _does_ appear then.

colin-kiegel:
@TedDriggs The skeptic tests have some issues with cargo cache. Since I activated both cargo cache and skeptic tests in master, I'm sure this failing build is just spurious nonsense and unrelated to this PR. The [failing test](https://travis-ci.org/colin-kiegel/rust-derive-builder/jobs/219789090#L644) looks at least similar to [this bug in skeptic](https://github.com/brson/rust-skeptic/issues/18). I'll look into a workaround - which might involve disabling skeptic tests again. :-)

I'll also take a look at your PR this weekend.

derive_builder/src/options/struct_mode.rs
```diff
@@ -70,14 +70,23 @@ impl OptionsBuilderMode for StructMode {
 
 impl From<OptionsBuilder<StructMode>> for (StructOptions, OptionsBuilder<FieldMode>) {
     fn from(b: OptionsBuilder<StructMode>) -> (StructOptions, OptionsBuilder<FieldMode>) {
+        let field_default = match b.default_expression {
+            Some(DefaultExpression::Trait)
```
TedDriggs: @colin-kiegel I think at this point you're right that DefaultExpression may be able to fall back to just having two variants: "Inherit" and "Explicit"

derive_builder/src/options/struct_options.rs
```diff
@@ -52,6 +56,20 @@ impl StructOptions {
...
         }
     }
 }
+
+/// Converts a DefaultExpression into the correct code block for the initializer.
+fn struct_default(expr: &DefaultExpression) -> Block {
+    (match *expr {
+        DefaultExpression::Struct 
+        | DefaultExpression::Trait => "::std::default::Default::default()".parse(),
+        DefaultExpression::Explicit(ref body) => format!("{}()", body).parse()
```
TedDriggs: Flagging this for closer scrutiny, as it touches on the issue described in #65 (this change is just at the struct level; the field level can be a separate PR).

derive_builder_core/src/build_method.rs
```diff
@@ -65,6 +68,11 @@ impl<'a> ToTokens for BuildMethod<'a> {
             BuilderPattern::Immutable => quote!(&self),
         };
         let doc_comment = &self.doc_comment;
+        let default = if let Some(ref expr) = self.struct_default { 
+            quote!(let __default : #target_ty = #expr;)
```
TedDriggs: Possible optimization: Check if all fields are explicitly initialized and avoid allocating the extra object if it won't be used.

derive_builder/src/options/struct_mode.rs
TedDriggs: Never mind; the initializer still needs some way of differentiating between "use the `Default` trait" and "use the `__default` object"

derive_builder_core/src/build_method.rs
TedDriggs: The only way I can see to do this is to generate an and-block calling `is_some()` on every field in the builder, and allowing `__default` to be `Option<#target_ty>` instead of `#target_ty`. I'd then have to call `unwrap` for every implicitly-instantiated field; I don't know if there's a good way to determine which path is more efficient.

colin-kiegel:
@TedDriggs I've started some minor refactoring on top of your branch - is it ok, if I push my changes to your branch, too? _Alternatively_ I could just push to a different branch and then maybe let you merge my changes into your PR - especially if you want to continue in parallel. However I'm not yet done. What do you think?

colin-kiegel:
@TedDriggs Btw. did you delete 2 comments? In my inbox I counted 5 inline-comments, but I count only three in your review - or am I missing something?

derive_builder_core/src/build_method.rs
colin-kiegel: I think I found a way to generate the code like this:
```rust
Foo {
    foo: 42,
    ..#struct_default // <-- only emitted if struct_default is set.
}
```

I will commit this as soon as I checked a few other things. :-)

colin-kiegel:
Hm, looks like something went wrong during rebase. I'll rebase again and probably force push.

colin-kiegel:
Ok, it turns out I actually overlooked something. My current approach does not work, if there is a struct default expression, but no field default expression. It looks like a temporary value is indeed unavoidable - I guess I will change it back, but not force push. :-)

# Current

```rust
    Foo {
            foo: match self.foo {
                Some(value) => value,
                None => /* ... _no_ field default, but we have to return something .... */,
            },
            ../* struct default expression */
    }
```

# Previous

```rust
    let __default = /* struct default expression */;
    Foo {
            foo: match self.foo {
                Some(value) => value,
                None => __default.foo,
            },
    }
```

Issue #25, opened 2016-08-26
Idea: Let the builder wrap a Result<T,Error>


nielsle:
Here is a loose idea. It is mostly meant as an inspiration..

It could be to have a builder that wraps a Result. This would allow the builder to validate fields during the build process, and keep track of errors. Rust already allows you to do this with the `try!()`, macro, and the ?-operator has been accepted as RFC 243, but I still think that it could be nice to have a builder that keeps track of errors.

Here is what the code could look like

```
// The following only works on unstable
![feature(try_from)]
use std::convert::{TryFrom, TryInto};

#[derive(Debug)]
enum  MyError {
    TooOld
}

// I am not fully sure if it makes sense to annotate the fields
// #[builder(PersonBuilder, MyError)]
#[derive(Default)]
struct Person {
    name: String,
    // #[try_from]
    age: Age,
    // #[try_from]
    mail: Mail
}

#[derive(Default)]
struct Age(u32);

impl TryFrom<u32> for Age {
    type Err = MyError;
    fn try_from(age: u32) -> Result<Age,MyError> {
        if age > 200 {
            Err(MyError::TooOld)
        } else {
            Ok(Age(age))
        }
    }
}

#[derive(Default)]
struct Mail(String);

// We should probably have TryFrom<&str> here
impl TryFrom<String> for Mail {
    type Err = MyError;
    fn try_from(mail: String) -> Result<Mail,MyError>{
        // Parsing goes here
        Ok(Mail(mail.into()))
    }
}

fn main() {

    let _person: Result<Person, MyError> = PersonBuilder::default()
        .name("bob".to_string())
        .age(25)
        .mail("a@b.c".to_string())
        .build();
}
```

Here is the code that could be derived behind the scenes

```
struct PersonBuilder {
    inner: Result<Person, MyError>
}

impl PersonBuilder {
    fn name(mut self, name: String) -> PersonBuilder {
        if let &mut  Ok(ref mut inner) = &mut self.inner {
            inner.name = name
        };
        self
    }

    // This code can probably be prettyfied
    fn age<T: TryInto<Age, Err=MyError>>(mut self, age: T)  -> PersonBuilder {
        match age.try_into() {
            Err(e) => if self.inner.is_ok() {
                self.inner = Err(e)
            },
            Ok(v) => if let &mut Ok(ref mut inner) = &mut self.inner {
                inner.age = v
            }
        };
        self
    }


[...]

PR #72, opened 2017-04-10
Add try_setter option and generate appropriate methods

TedDriggs:
This implements the method-level proposal described in #25. The initial PR is not ready for merging; it's meant to start a discussion.

# Open Questions
1. What blanket `TryFrom` and `TryInto` implementations will be provided when the feature stabilizes? Right now, the lack of support among built-in types makes this only narrowly useful. It looks like there will be a blanket impl for `TryFrom<&'a str>` for things that currently implement `FromStr` ([source comment](https://github.com/rust-lang/rust/issues/33417#issuecomment-277253827))
1. Is there an annotation needed on the generated methods to allow an exporting crate to offer these methods only when the consuming crate has enabled the language `try_from` feature?


colin-kiegel:
Hey, that was fast. Hm, good questions - especially the second one.
- String parsing will probably be the most ubiquitous use case.
- I think [`#[cfg_attr(feature = "foo", builder(...))]`](https://chrismorgan.info/blog/rust-cfg_attr.html) should allow crates to set builder attributes conditionally. This could be added to a tips section. And if that doesn't work I'd just wait and hope that `TryFrom` doesn't take too long for stabilization.

I'll look at the code tomorrow. :-)

derive_builder_core/src/setter.rs
```diff
@@ -117,6 +119,27 @@ impl<'a> ToTokens for Setter<'a> {
...
+                let try_ty_params = quote!(<VALUE: #try_into<#ty>>);
+                let try_ident = syn::Ident::new(format!("try_{}", ident));
+                let result = self.bindings.result_ty();
+                
+                tokens.append(quote!(
+                    #(#attrs)*
+                    #vis fn #try_ident #try_ty_params (#self_param, value: VALUE)
+                        -> #result<#return_ty, VALUE::Err>
+                    {
+                        #deprecation_notes
```
colin-kiegel: Please remove this line, because we want to emit this only _once_ per setter combination.
Since we piggy back the try-variant onto the standard setter, it will already be emitted in the standard setter.

derive_builder/Cargo.toml
```diff
@@ -25,6 +25,7 @@ proc-macro = true
 logging = [ "log", "env_logger", "derive_builder_core/logging" ]
 struct_default = []
 skeptic_tests = ["skeptic"]
+try_setter = []
```
colin-kiegel: The user should not need to set a feature flag to use generate try-setters.

Instead we only need some feature flag to activate the integration tests. The situation with having multiple crates in a workspace and running tests on travis is a bit complicated. When you look at [`travis.yml:57`](https://github.com/colin-kiegel/rust-derive-builder/blob/e44928318ea7f47f8f7483ce335fbd6c853024f5/.travis.yml#L57), you'll see that tests are started in the `derive_builder_test` folder with the arguments `--all --no-fail-fast --features "$CARGO_FEATURES"`. My experience is that the features are **only** passed to the `derive_builder_test` crate, but not `derive_builder` or `derive_builder_core` unless `derive_builder_test/Cargo.toml` specifies to _propagate_ the features (which is does not right now). So the easiest solution is to place all integration tests that require nightly into `derive_builder_test/tests` and use a feature flag in the `derive_builder_test` crate. That also has the advantage that the public `derive_builder` does not expose any feature flag that's useless for users.

You can use the [`compiletests`](https://github.com/colin-kiegel/rust-derive-builder/blob/e44928318ea7f47f8f7483ce335fbd6c853024f5/.travis.yml#L12-L13) feature flag and rename it to `nightlytests`. Please also change all occurences of `feature = "compiletests"`, but don't change filenames `compiletests.rs` or other occurences of the word. ;-)

Let me know if you need any further advice.

derive_builder/src/options/mod.rs
```diff
@@ -198,6 +206,9 @@ impl<Mode> OptionsBuilder<Mode> where
                 // setter implicitly enabled
                 self.setter_enabled(true)
             },
+            "try_setter" => {
+                self.try_setter_enabled(true)
```
colin-kiegel: please set `setter_enabled` implicitly if the user specifies `#[builder(try_setter)]`
```rust
"try_setter" => {
    /* ... */
    // setter implicitly enabled
    self.setter_enabled(true)
}
```

colin-kiegel: Or wait - let me think about it again. Initially I thought you were implementing an API like `#[builder(setter(try))]` as in [this comment](https://github.com/colin-kiegel/rust-derive-builder/issues/25#issuecomment-293044392).

colin-kiegel:
@TedDriggs before we continue with this PR let's first discuss what behaviour want to have in [issue #25](https://github.com/colin-kiegel/rust-derive-builder/issues/25#issuecomment-293227906)

derive_builder/src/options/mod.rs
TedDriggs: I think the `#[builder(setter(try))] is the better API in the long run, as having a `try_{field}` without the regular setter seems bad.

colin-kiegel:
I was thinking, if we go with the approach (c) discussed [here](https://github.com/colin-kiegel/rust-derive-builder/issues/25#issuecomment-293227906), it might make sense to abstract over `try_setter` vs. `setter` differently. The current implementation of a `Setter` object can emit both setters in one go. Alternatively we could have to methods `as_setter()` and `as_try_setter()` on `FieldOptions`. Both would then each generate one `Setter` object with different settings. This would keep the complexity in the `Setter` object lower when we add options to overwrite the `try_setter` name etc.

I'm absolutely ok with deferring that refactoring to the point where we need it. But if you feel like doing it now, I'd welcome that of course. You decide. ;-)

derive_builder/src/options/mod.rs
colin-kiegel: ok, please forget that comment about implicitly enabling the standard setter - as we decided _not_ to piggy back. :-)

derive_builder_core/src/setter.rs
```diff
@@ -58,7 +60,7 @@ pub struct Setter<'a> {
 
 impl<'a> ToTokens for Setter<'a> {
     fn to_tokens(&self, tokens: &mut Tokens) {
-        if self.enabled {
+        if self.enabled || self.try_enabled {
```
colin-kiegel: PS: You can either pull all shared variables out of the `if self.enabled {` statement, or just emit two `Setter` objects per field (as suggested in another comment).

TedDriggs:
There's one more issue I need to address, which is controlling the generation of these based on the presence of the `try_from` language feature. There appear to be three possible ways to accomplish this:

1. Do a `cfg!(feature = "try_from")` check in `derive_builder`, and pass `false` to core if that feature isn't enabled.
1. Do a `cfg!(feature = "try_from")` check in `derive_builder_core`, and only emit the setter when the language feature is present.
1. Emit `#[cfg(feature = "try_trom")]` on the generated try_setter. This might require reconciling with existing `cfg` attributes added on the base struct.

I think approach 1 is the simplest, but I'm not sure if that makes the determination about exposing these be part of the _exporting_ crate, or the _consuming_ crate. Ideally, the consuming crate's language features would control this.

derive_builder_core/src/setter.rs
```diff
@@ -256,14 +256,36 @@ mod tests {
         setter.attrs = attrs.as_slice();
         setter.generic_into = true;
         setter.deprecation_notes = &deprecated;
+        setter.try_setter = true;
+
+        // Had to hoist these out to avoid a recursion limit in the quote!
+        // macro invocation below.
```
colin-kiegel: no, problem. You can just add `#![cfg_attr(test, recursion_limit="100")]` to the crate root. That should increase the recursion limit in test-mode only. Just put in a number that's roughly twice as high as the current minimum requirement. :-)

derive_builder_core/src/setter.rs
```diff
@@ -305,4 +327,36 @@ mod tests {
...
+        
+        let setter_quote = quote!(
+            pub fn foo(&mut self, value: Foo) -> &mut Self {
+                let mut new = self;
+                new.foo = ::std::option::Option::Some(value);
+                new
+            }
+        );
+        
+        // hoisting necessary to avoid recursion limit.
```
colin-kiegel: same here

derive_builder_test/tests/run-pass/try_setter.rs
```diff
@@ -1,11 +1,11 @@
-#![cfg_attr(feature = "try_setter", feature(try_from))]
+#![cfg_attr(feature = "nightlytests", feature(try_from))]
 
 #[macro_use]
 extern crate derive_builder;
 
 #[allow(unused_imports)]
 mod struct_level {
-    #[cfg(feature = "try_setter")]
+    #[cfg(feature = "try_from")]
```
colin-kiegel: ok, gating on `try_from` instead of `nightlytests` is actually a good idea here. Please do it consistently in the rest of the test then - that'd be easier to understand. :-)

Alternatively also gate on `nightlytests` here.

derive_builder/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
 
+## [0.4.3] - 2017-04-13
+
+### Added
+- try_setters, e.g. `#[builder(try_setter)]`. These setters are exposed alongside the
+  normal field setters and allow callers to pass in values which have fallible 
+  conversions to the needed type through `TryInto`. This attribute can be used on any 
+  Rust channel, but will only emit setters on nightly when `#![feature(try_from)]` is 
+  declared in the consuming crate's root; this will change when Rust issue 
+  [#33417](https://github.com/rust-lang/rust/issues/33417) is resolved.
```
colin-kiegel: nice summary! :+1:

derive_builder/src/lib.rs
```diff
@@ -366,6 +391,9 @@
 //! - If derive_builder depends on your crate, and vice versa, then a cyclic
 //!   dependency would occur. To break it you could try to depend on the
 //!   [`derive_builder_core`] crate instead.
+//! - The `try_setter` attribute and `owned` builder pattern are not compatible in practice;
+//!   an error during building will consume the builder, making it impossible to continue
+//!   construction.
```
colin-kiegel: nice catch :-)

derive_builder/src/lib.rs
```diff
@@ -216,6 +216,31 @@
...
+//! # extern crate derive_builder;
+//! #
+//! #[derive(Builder, Debug, PartialEq)]
+//! #[builder(try_setter, setter(into))]
+//! struct Lorem {
+//!     pub ipsum: u8,
+//! }
+//! ```
+//!
+//! A complete example of using fallible setters is available in `examples/try_setter.rs`.
```
colin-kiegel: please add a note that the name and visibility are inherited from the `setter` settings.

You could change the example into something like this, to better explain it?
```rust
//! # #[macro_use]
//! # extern crate derive_builder;
//! #
//! #[derive(Builder, Debug, PartialEq)]
//! struct Lorem {
//!     #[builder(try_setter, setter(name="foo"))]
//!     pub ipsum: u8,
//! }
```

derive_builder/examples/try_setter.rs
```diff
@@ -0,0 +1,52 @@
...
+    pub addr: MyAddr,
+}
+
+fn main() {
+    create("Jane", "1.2.3.4").unwrap();
+    create("Bobby", "").unwrap_err();
+}
+
+fn create(name: &str, addr: &str) -> Result<Lorem, String> {
+    // Fallible and infallible setters can be mixed freely when using
```
TedDriggs: @colin-kiegel can you please try running `cargo expand --example try_setter` on this and watch what happens to the comment? Something weird is happening here, but I'm not sure what.

colin-kiegel:
> [There's one more issue I need to address, which is controlling the generation of these based on the presence of the try_from language feature. There appear to be three possible ways to accomplish this ...](https://github.com/colin-kiegel/rust-derive-builder/pull/72#issuecomment-293330211)

I think all three variants don't work, because I ran into **exactly** the same problem with `no_std` support in #45. TL;DR: I decided to leave it to the user of `derive_builder` to manually set `#[builder(no_std)]`. It was _not_ possible to detect whether the user set `#![no_std]`. I suggest to do the same here. :-)

derive_builder/examples/try_setter.rs
colin-kiegel: Let me guess, does the comment appear at _random_ places in the generated output? :-)

I noticed _very_ strange distributions of comments before. I could imagine that this is caused by the syn crate, which we use to parse the input into an ast. But didn't investigate it, because it didn't cause any trouble.

derive_builder/CHANGELOG.md
```diff
@@ -6,10 +6,10 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - try_setters, e.g. `#[builder(try_setter)]`. These setters are exposed alongside the
-  normal field setters and allow callers to pass in values which have fallible 
-  conversions to the needed type through `TryInto`. This attribute can be used on any 
-  Rust channel, but will only emit setters on nightly when `#![feature(try_from)]` is 
-  declared in the consuming crate's root; this will change when Rust issue 
+  normal field setters and allow callers to pass in values which have 
+  fallible conversions to the needed type through `TryInto`. This attribute 
+  can only be used on nightly when until `#![feature(try_from)]` is declared
```
colin-kiegel: `when until` -> `when`

TedDriggs:
> can you revive the tests you had in the old version of derive_builder_test/tests/try_setter.rs?

Done, and added a more complete `Ipsum` to get coverage on renaming.

> PS: Are you developing on Linux? I can commit some local shell scripts that I use to emulate what travis is doing. I find them really useful to test all the different combinations nightly with features, stable / beta without..

OS X; I can try the scripts to see if they'll work for me.

colin-kiegel:
ok, I pushed a [devtools](https://github.com/colin-kiegel/rust-derive-builder/tree/devtools/dev) branch. There is a `dev` folder and a `README.md` which hopefully explains everything. Let me know if that works for you - if it does, I'll merge it to master.

colin-kiegel:
PS: Note running the tests might take a little while, because this will call `rustup update` for each toolchain. It buffers all writes to stdout/stderr and only prints it on failure. A successful run should look like this:
```
II: Working directory is... ✓ clean
II: Running tests on stable... ✓
II: Running tests on beta... ✓
II: Running tests on nightly... ✓
II: Running dev/compiletests.sh... ✓
II: Running dev/checkfeatures.sh... ✓
OK: All checks passed!
```

derive_builder/examples/try_setter.rs
TedDriggs: It seems to consistently get moved into the `build` method just before the uninitialized field error string for "name", even when I move it around in the raw source.

TedDriggs:
@colin-kiegel I've pulled down that branch, and found one _an_ issue. It appears OS X doesn't support `readlink -f`, which is causing a bunch of failures. I'm far from a bash expert, but I got it working with this:

```bash
function base_dir {
  if [ uname -s != "Darwin" ] -a hash readlink 2>/dev/null; then
    # use readlink, if installed, to follow symlinks
    local __DIR="$(dirname "$(readlink -f "$0")")"
  else
    local __DIR="$(dirname "$0")"
  fi
  echo ${__DIR}
}
```

OS X appears to reliably return `Darwin` for the uname, which was convenient. After I made that change, running `dev/githook.sh` started passing all tests, but when they ran as part of the push process they failed saying they couldn't find the `derive_builder_test` directory. That's probably related to my edit, but I don't know any Bash so I ran out of ability to diagnose after that.

colin-kiegel:
The basedir() function is supposed to get the directory of the script independently of how it was called. It may seem like overkill to follow symlinks, but actually that's what went wrong, because `.git/hooks/pre-push` is a symlink. Without following symlinks, the directory of the symlink will be reported by basedir().

To workaround this, you can remove the symlink in the `.git` directory and replace it with a wrapper bash script `.git/hooks/pre-push`
```bash
#!/bin/bash
../../dev/githook.rs
```
Or search for a way how to follow symlinks on OS X.

colin-kiegel:
It's great that this was the only issue! :-)

Note the skeptic and compile tests sometimes have linking problems, when switching branches, updating dependencies etc. In that case just run `cargo clean`.

TedDriggs:
Okay, so this is interesting... the one that fails says `TryFrom` doesn't have an associated type `Err`, then later complains that the associated type `Err` is missing. Not sure how to interpret that issue.

colin-kiegel:
I'd `cargo expand` a simple example and copy that into another​ temporary example file and debug from there.

TedDriggs:
Or once I make that change everything will work :)

PR #74, opened 2017-04-11
Fix #68 - setter(skip) with struct and field-explicit defaults

### LINKED ISSUE #68: setter(skip) should use the custom default if that was defined
If a field is annotated with `#[builder(setter(skip))]` then the `Default` trait will be used to construct a value.

The same field (or even the whole struct) can be annotated with an explicit default `#[builder(default="..."))]`. If that is the case, this should be used to construct the value. This is currently not the case and a bug.

TedDriggs:
Includes updated tests and changelog entry. This fixes issue #68.

### DIFF
--- a/derive_builder/CHANGELOG.md
+++ b/derive_builder/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.4.3] - 2017-04-11
+
+### Fixed
+- `setter(skip)` honors struct-inherited and explicit defaults #68
+
 ## [0.4.2] - 2017-04-10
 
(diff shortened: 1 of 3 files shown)


TedDriggs:
Just realized, at least one of the tests depends on `struct_default`; working on a fix for that.

derive_builder/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.4.3] - 2017-04-11
```
colin-kiegel: You can just add `## [Unreleased]` at the top in the future - at least that's how I do it. I've got that workflow from http://keepachangelog.com/. ;-)

Issue #76, opened 2017-04-12
Implement TryFrom<Builder> for Target for all builders exposing a public build method


TedDriggs:
With #72 (fallible setters) merged and rust-lang/rust#33417 (`TryFrom`) approaching stabilization, having a `TryFrom<Builder> for Target` implementation for all builders with a generated `build` method will enable the following code:

```rust
let request = RequestBuilder::default()
    .from(-18000000)
    .sources(vec![Source::device_group(Oid::new(2))])
    .metric(Metric::builder()
        .stat_name("extrahop.device.ssl_server_detail")
        .field_name("cert_expiration"))?
    .total(true)
    .build()
    .expect("All inputs hardcoded");
```

Once `TryFrom` is stabilized I'm not planning to expose a struct-level attribute to control this. This seems too standard for a crate user _not_ to want. #70 gives authors control over exposing the `build` method; this would look to that attribute and skip implementation if the `build` method was skipped.

I'm planning to take a stab at a PR for this.

colin-kiegel:
Makes sense :+1:

I generally agree that we should always emit that implementation. There is only one reason to make it opt-into and that is backwards compatibility with Rust 1.15-x. I don't want to drop backwards compatibility on every new cool feature. For some intermediate time I would like to make it opt-in (either feature flag or attribute). And after some time we can raise the minimum Rust version and get rid of opting-in. I guess a feature flag would be a little easier to implement? With this we can already implement it for nightly. :-)

Minor corrections in your code example
* You meant `RequestEnvelope::try_from`
* the `build().expext("..")` calls should not be there
* `Metric::builder()` would _currently_ still be `MetricBuilder::default()`

TedDriggs:
* The `expect` call should unwrap the build result with a nicer panic message. In this case, there was no `RequestEnvelopeBuilder`, but there was a hand-rolled `From` impl.
* See point 1
* Oops; I've added that by hand in the crate where this was copy-pasted from. #73 

I went with feature flag.

colin-kiegel:
Ah .. I see .. I thought the outer `RequestEnvelope::from(RequestBuilder` should've _also_ been a demonstration of the usefulness of the automatic `TryFrom` implementation.

Ok, that was fast. I'll review later. Have to got now. :-)

TedDriggs:
`TryFrom` is now supported below this crate's MSRV, so this should be ready to implement.

MrBeeMovie:
[...]

Issue #54, opened 2017-03-20
Change default build_fn name from `build` to `try_into` in preparation for `TryFrom`/`TryInto` traits?


colin-kiegel:
[`TryFrom<T>` and `TryInto<T>`](https://github.com/rust-lang/rfcs/blob/master/text/1542-try-from.md) are about to be stabilized. These would be the idiomatic builder traits IMO. And they are likely to end up in the std prelude, so you _wouldn't_ need to import them.

The current build method, could be replaced by `TryFrom` (which implies `TryInto`)
```rust
impl TryFrom<FooBuilder>: Foo {
    type Err = String;

    fn try_from(b: FooBuilder) -> Result<Foo, String> {
         // ...
    }
}
```

_Removing_ `FooBuilder::build()` in favour of `TryFrom::try_from()` would be a breaking change. 

However we could keep the current default fn `build()` until `TryFrom` is stabilised. Then let `build` wrap around `TryFrom` and keep it this way virtually forever (unless opted out).

Alternatively, we could try to get rid of the wrapping `build()` fn as early as possible. This could happen in two steps (a) rename `build()` to `try_into()` and (b) switch to `TryFrom`, while `builder.try_into()` should still magically work  in almost all cases.  There could still be an option to opt into creating such a wrapper function with a custom name.

`derive_builder` is still young, so we may deprecate things more easily now rather than later. The question is, where do we want to end up long term? Do we want to keep a `build()` fn wrapper as default behaviour, or do we want to advocate using `TryFrom` and not have a wrapper by default?

TedDriggs:
The `try_into` method would be both inherent and trait-defined, right? Having to add another `use` statement would be unfortunate. Also, I vaguely recall something about offering descriptive inherent method names, but I can't find it on Google and it may have been specific to types in `std`.

colin-kiegel:
It depends. If `use TryFrom; use TryInto` were encluded in the std prelude, I could imagine only offering these trait implementations by default. An inherent / type method could then be generated by demand.

If these traits don't end up in the std prelude, then an inherent method is definitely needed.

Speaking of std: Generally I would like to mimic design philosophies of `std` as close as possible, to make this crate intuitive. So if you think that something violates a design principle of std that would be a strong argument for me to reconsider these choices. :-)

[...]

PR #77, opened 2017-04-12
Added TryFrom impl generation for target types behind feature flag

TedDriggs:
This addresses #76. The generation is automatic but depends on crate feature `try_from`.


TedDriggs:
@colin-kiegel The failed test looks to be related to a failure in rustup; don't know what to make of that.

colin-kiegel:
That's https://github.com/rust-lang-nursery/rustup.rs/issues/1062 and should be gone with the next nightly release. I can hit `restart` over the weekend and also have a look at the PR. :-)

derive_builder/Cargo.toml
```diff
@@ -23,6 +23,7 @@ proc-macro = true
 
 [features]
 logging = [ "log", "env_logger", "derive_builder_core/logging" ]
+try_from = []
 struct_default = []
 skeptic_tests = ["skeptic"]
 nightlytests = []
```
colin-kiegel: let's change this to `nightlytests = ["try_from"]` - this will in turn activate it in travis and `dev/nightlytests.sh` / `dev/githook.sh`.

derive_builder_core/src/build_method.rs
```diff
@@ -67,11 +67,7 @@ impl<'a> ToTokens for BuildMethod<'a> {
         let target_ty = &self.target_ty;
         let target_ty_generics = &self.target_ty_generics;
         let initializers = &self.initializers;
-        let self_param = match self.pattern {
-            BuilderPattern::Owned => quote!(self),
-            BuilderPattern::Mutable |
-            BuilderPattern::Immutable => quote!(&self),
-        };
+        let self_param = self.pattern.to_build_method_tokens();
```
colin-kiegel: I'd like to keep this match statement in `BuildMethod` for the moment. I feel like from the perspective of unit testing this gives clearer boundaries of responsibility. Of course this makes the `BuildMethod` type a little more complex..

derive_builder_core/src/options.rs
```diff
@@ -19,9 +21,20 @@ pub enum BuilderPattern {
...
+    /// Expresses this pattern
+    pub fn to_build_method_tokens(&self) -> Tokens {
+        match *self {
+            BuilderPattern::Owned => quote!(self),
+            BuilderPattern::Mutable |
+            BuilderPattern::Immutable => quote!(&self),
+        }
+    }
+}
+
```
colin-kiegel: please move this back to `BuildMethod` for the moment.

derive_builder_core/src/try_from.rs
```diff
@@ -0,0 +1,174 @@
...
+        let builder_ty = &self.builder_ty;
+        let target_ty = &self.target_ty;
+        let (owned_generics, ty_generics, where_clause) = self.generics.split_for_impl();
+        let fn_ident = &self.fn_ident;
+        let try_from = self.bindings.try_from_trait();
+        let result_ty = self.bindings.result_ty();
+        let string_ty = self.bindings.string_ty();
+        
+        let immutable_borrow = quote!(
+            impl #impl_generics #try_from<&'unused #builder_ty #ty_generics> for #target_ty #ty_generics
```
colin-kiegel: What happens if we elide the lifetime parameter `'unused` and write `& #builder_ty #ty_generics` instead?
Would that fail if `#ty_generics` contained other lifetime parameters? Did you test that?

If we really need to have an explicit lifetime parameter, I'd suggest to call it `'try_from` just in case so the user would get nicer error messages - while still reducing the risk of name collisions.

derive_builder_core/src/try_from.rs
```diff
@@ -0,0 +1,174 @@
...
+        let try_from = self.bindings.try_from_trait();
+        let result_ty = self.bindings.result_ty();
+        let string_ty = self.bindings.string_ty();
+        
+        let immutable_borrow = quote!(
+            impl #impl_generics #try_from<&'unused #builder_ty #ty_generics> for #target_ty #ty_generics 
+                #where_clause {
+                type Error = #string_ty;
+                
+                fn try_from(v: &#builder_ty #ty_generics) -> #result_ty<Self, Self::Error> {
```
colin-kiegel: you don't have a lifetime parameter here, that makes me suspicious that we might not need it. :-)

derive_builder_core/src/try_from.rs
```diff
@@ -0,0 +1,174 @@
...
+                }
+            }
+        );
+        
+        tokens.append(quote!(
+            #immutable_borrow
+            
+            #mutable_borrow
+            
+            #owned
```
colin-kiegel: I'd like to start with only one impl depending on the builder pattern

# Variant 1: Mimic `build()` method

Pseudo code for demonstration:

```rust
Owned => /* ... impl TryFrom<FooBuilder> for Foo ... */,
Mutable |
Immutable => /* ... impl TryFrom<&FooBuilder> for Foo ... */,
```

So despite having three variants of `BuilderPattern` there would only be two variants of `TryFrom` implementations - in analogy to the `BuildMethod`.

# Variant 2: Always take `Self`

The logic behind changing the signature of the build method was to enable method chaining `FooBuilder::default().foo(1).build()`. But this would most likely not be the primary use case of `TryFrom` - although it _could_ be used in this position.

Alternatively we could talk about _always_ implementing `TryFrom<FooBuilder> for Foo` and _never_ implementing `TryFrom<&FooBuilder> for Foo`.

# Discussion

We should probably discuss the primary use cases for having `TryFrom` implementations. What signature would they require? Would autoderef or other coercions help to pick the right `TryFrom` implementation?

derive_builder_test/Cargo.toml
```diff
@@ -9,7 +9,7 @@ publish = false
 path = "src/lib.rs"
 
 [features]
-nightlytests = ["compiletest_rs", "derive_builder/nightlytests"]
+nightlytests = ["compiletest_rs", "derive_builder/nightlytests", "derive_builder/try_from"]
```
colin-kiegel: I removed the `derive_builder_test` crate in latest master. :-)

PS: More correctly - I merged it into the `derive_builder` crate and left a tiny `derive_builder/tests/dummy-crate` for doc-inspection.

derive_builder_test/tests/try_from.rs
```diff
@@ -0,0 +1,37 @@
+#![cfg(feature = "nightlytests")]
+#![cfg_attr(feature = "nightlytests", feature(try_from))]
```
colin-kiegel: please merge (or rebase to) current master. Then depend on `feature = "try_from"` instead.

TedDriggs:
@colin-kiegel are the run-pass tests not passed the flags from the crate?

colin-kiegel:
@TedDriggs Hm, it would seem so. You could try something like this for the compiletests. 

```diff
-#![cfg_attr(feature = "nightlytests", feature(try_from))]
+// Compiletests seem to _not_ have access to feature flags. We need to set `#![feature(try_from)]` unconditionally.
+#![allow(unused_features)]
+#![feature(try_from)]
```

TedDriggs:
Now it's complaining about the example in `readme.md`. I'm stumped on how to get this working without a separate feature flag gating the generation of the `TryFrom` blocks.

colin-kiegel:
I imagine the final `TryFrom` implementation like this:
- the `build`-method will only be a shallow wrapper around the `TryFrom` implementation
  - all logic (like custom validation) should be implemented in the `TryFrom` implementation
  - the build method will require the `TryFrom` implementation - not the other way around, like now
  - the build_fn will follow the builder pattern and clone when necessary, before delegating to `TryFrom`.
  - the `TryFrom` implementation always consumes `FooBuilder` and avoids cloning. (I believe this would be the most idiomatic implementation)

It's not a problem to start with `TryFrom` delegating to the build fn (i.e. in the other direction), before we remove the feature gate. However IMO we should already do the following change in this PR:
- `TryFrom` should always consume `FooBuilder` (i.e. not take references)

What do you think about this?

colin-kiegel:
PS: If you happen to agree to this change than IMO we would be ready to merge. Otherwise we can of course discuss this more deeply. :-)

TedDriggs:
> the build-method will only be a shallow wrapper around the TryFrom implementation

My impression after #54 was that logic should live in inherent methods, with the trait impl being the shallow wrapper.

> TryFrom should always consume FooBuilder (i.e. not take references)

Without the version that takes a reference, using this in conjunction with #72 gets more verbose. The setter methods are returning `&mut Self` by default, so not having `impl<'a> TryFrom<&'a mut FooBuilder> for Foo` means you can't write:

```rust
let bar = BarBuilder::default()
    .try_baz("1.2.3.4")?
    .try_foo(FooBuilder::default()
        .lorem("Hello")
        .ipsum("world"))?    
    .build()?;
```

I'm not opposed to having an owned, non-cloning implementation, as long as Rust is smart enough to avoid an extra clone to push into that implementation if someone writes the code above.

colin-kiegel:
> My impression after #54 was that logic should live in inherent methods, with the trait impl being the shallow wrapper.
Hm, but it does not make a difference with regard to inference whether a trait impl delegates to an inherent method or vice-versa. The main issue in #54 was that things like `.try_into().try_into().try_into()` have inference problems. But inherent methods `.foo().bar().baz()` do not have this problem, even if each one of them delegates to a trait impl.

> The setter methods are returning &mut Self by default
I'm actually _thinking_ about changing that default to `Self`. I started a discussion #88 about this - but I don't intend to rush anything there, because we already did a flip in the past.

colin-kiegel:
It's already possible to use nested builders even with the mutable pattern. You only need to write `.foo(foo_builder.bar(v).build()?)` instead of `.try_foo(foo_builder.bar(v))?`.

I think we can still introduce the `TryFrom<&FooBuilder>` implementation later. It would be less of a breaking change if we introduce it later rather than remove it later.

Which reminds me that we should raise the minor version with this feature.

PR #78, opened 2017-04-13
Documentation-related enhancements

TedDriggs:
Two minor things (both vaguely docs-related).

1. Made the builder's documentation include a link to the target type. Since we know we're in the same module as the target type, we can do this reliably with a relative URL.
1. Added a module-level doc comment to `derive_builder::options` to help others understand what's happening. @faern, let me know if this helps.

### DIFF
--- a/derive_builder/src/doc_tpl/builder_struct.md
+++ b/derive_builder/src/doc_tpl/builder_struct.md
@@ -1,1 +1,1 @@
-Builder for `{struct_name}`.
+Builder for [`{struct_name}`](struct.{struct_name}.html).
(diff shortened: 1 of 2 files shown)


derive_builder/src/doc_tpl/builder_struct.md
```diff
@@ -1 +1 @@
-Builder for `{struct_name}`.
+Builder for [`{struct_name}`](struct.{struct_name}.html).
```
colin-kiegel: Note: Both the name and the link will break, if `{struct_name}` is re-exported, because the generated documentation will not know about the re-export. This is something we cannot fix - at least not automatically.

To see what I mean you can run this on current master
* `cd derive_builder/tests/dummy-crate`
* `cargo doc`

Issue #67, opened 2017-04-09
Custom validation of built struct


faern:
It would be nice to allow custom code to validate an object before it is returned from the `build` method. This topic is touched upon in #53 and #56, but none of them are really about this, nor discuss it in detail.

I mean to validate the entire struct upon build, not individual fields on set. This is for situations where there are no invalid values for individual fields, but some combinations of field values are invalid.

## Example
```rust
#[derive(Builder)]
#[builder(validatefn=my_validate)]
pub struct Foo {
    src: IpAddr,
    dst: IpAddr,
}

impl FooBuilder {
    fn my_validate(foo: &Foo) -> Result<(), String> {
        if foo.src.is_ipv4() == foo.dst.is_ipv4() {
            Ok(())
        } else {
            Err("Source and destination addresses not in same address family".to_owned());
        }
    }
}

// And the generated builder looks something like:
impl FooBuilder {
    fn build(&mut self) -> Result<Foo, String> {
        let foo: Result<Foo, String> = whatever_it_does_now_to_create_foo();
        foo.and_then(|foo| Self::my_validate(&foo).map(|_| foo))
    }
}
```

The value sent to `validatefn` should probably be limited to an identifier, a method name, and the code generated will be `Self::$validatefn(&value)...`

## Structured errors

Structured errors, instead of `String` is discussed in issue #60. I really like the idea of returning structured errors and support that. This custom validation can be made to support that by changing the requirement on `validatefn` from (`T` is the type we are building):
```rust
Fn(&T) -> Result<(), String>
```
to
```rust
Fn(&T) -> Result<(), E> where E: std::error::Error
```
And introduce another structured error variant `ValidationFailed(E)`, giving us a `build` method similar to:
```rust
impl FooBuilder {
    fn build(&mut self) -> Result<Foo, FooBuilderError> {
        let foo: Result<Foo, FooBuilderError> = whatever_it_does_now_to_create_foo();
        foo.and_then(|foo| {
            Self::my_validate(&foo)
                .map(|_| foo)
                .map_err(|e| FooBuilderError::ValidationFailed(e))
        })
    }
}
```
Allowing the user of `derive_builder` to return any error type they want in their validation function.

faern:
Should just add that I'm willing to help build this. But I want to see how it is received first.

faern:
[...]

PR #79, opened 2017-04-14
Vision proposal

TedDriggs:
Here's the vision proposal doc related to discussion of #67. @colin-kiegel and @faern, you should now be able to add inline comments to the proposal.

EDIT: [Rendered](https://github.com/TedDriggs/rust-derive-builder/blob/vision/vision_doc.md).


vision_doc.md
```diff
@@ -0,0 +1,279 @@
...
+        }
+    }
+    
+    impl TryFrom<StatRequestBuilder> for StatRequest {
+        type Error = String;
+        
+        fn try_from(v: StatRequestBuilder) -> Result<Self, Self::Error> {
+            v.build()
+        }
+    }
```
TedDriggs: The `TryFrom` could be automatically generated if the user told us that the build method was implemented and what its name was, but I take your point. I was also trying to avoid any magic that depended on field order being the same in the struct declaration and the constructor, although something could be done with that which would make the code shorter.

I tried to pick an example where the validation logic was non-trivial and would need to be applied regardless of how the object was assembled.

vision_doc.md
```diff
@@ -0,0 +1,279 @@
...
+
+The builder should be focused specifically on the assembly of inputs and validating the completeness of the parameter set; additional validation should be handled by a constructor function provided by the target type. That constructor function may be public or private, depending on whether or not the crate author wants to _require_ the use of the builder.
+
+## Documentation
+Authors should focus documentation efforts on the target type.
+* If the builder is the only way to create the target type, then the target type should discuss the requirements for its own creation.
+* The builder can automatically include a link to the target type doc HTML from its own struct-level doc comment.
+
+## Imports
+A crate/module should always export the target type of a builder to allow the _built_ type to appear in function and struct type declarations. The exporting crate should _generally_ also expose the builder type, but may choose to exclude it from a prelude or the crate root, preferring to expose it in a child module. The crate is also free to keep the builder for its own internal use.
```
TedDriggs: If _only_ `FooBuilder` is exposed, then the importing crate has no way to refer to the constructed type in structs or function signatures. I can't think of a case where that's desirable.

vision_doc.md
```diff
@@ -0,0 +1,279 @@
+# Vision
+`derive_builder` provides an ergonomic way to deal with construction of guaranteed-complete objects from a range of inputs:
+
+* Statically-known values
+* Function arguments
+* Wire/file/database inputs
+
+The builder should be focused specifically on the assembly of inputs and validating the completeness of the parameter set; additional validation should be handled by a constructor function provided by the target type. That constructor function may be public or private, depending on whether or not the crate author wants to _require_ the use of the builder.
```
TedDriggs: I've moved it to a sub-section, but I do think that having some level of opinion on where validation fits in is significant. (Also, not sure where that statement would go otherwise as the doc stands now).

vision_doc.md
```diff
@@ -0,0 +1,279 @@
+# Vision
```
TedDriggs: Done.

vision_doc.md
```diff
@@ -0,0 +1,279 @@
...
+* The builder can automatically include a link to the target type doc HTML from its own struct-level doc comment.
+
+## Imports
+A crate/module should always export the target type of a builder to allow the _built_ type to appear in function and struct type declarations. The exporting crate should _generally_ also expose the builder type, but may choose to exclude it from a prelude or the crate root, preferring to expose it in a child module. The crate is also free to keep the builder for its own internal use.
+
+For the case of builder population with statically-known values or through explicit construction from function arguments, the target type should (optionally) expose a static `builder()` method which returns the builder type. In addition to keeping the imports shorter, this will appear in RLS "find all reference" queries by type, will be updated automatically during type renaming within the workspace, and will show docs for the target type on type hover (requesting docs on the static method will get the user to the builder-specific documentation).
```
TedDriggs: Added links to the relevant RLS methods.

vision_doc.md
```diff
@@ -0,0 +1,279 @@
...
+    /// Could be private if the builder was meant to be only way to create instance.
+    pub fn new(host: String, api_key: ApiKey, timeout: Duration) -> Result<Self> {
+        // @FAERN; validation would occur here
+        
+        let socket = unimplemented!(); // try to open the socket...
+        ConnectionPool {
+            host: host,
+            api_key: api_key,
+            timeout: timeout
+        }
```
TedDriggs: Fixed.

vision_doc.md
```diff
@@ -0,0 +1,279 @@
...
+    /// 
+    /// * `build_fn(skip)`: This tells the builder not to auto-generate the 
+    ///   build method.
+    /// * `preserve_attributes`: This tells the builder to include all #[serde] 
+    ///   attributes on the `StatRequestBuilder` fields.
+    /// * `target_factory`: Name needs work, but this tells the macro to add an 
+    ///   inherent `builder` function on the target type.
+    #[derive(Debug, Clone, Builder, Serialize)]
+    #[builder(setter(into), try_setter, preserve_attrs(serde), derive = "Deserialize", build_fn(skip), target_factory)]
+    pub struct StatRequest {
```
TedDriggs: Is your concern the `preserve_attrs(serde)` part, or the whole thing?

vision_doc.md
faern: A possible usecase could be where you always just send the constructed value back into the library. A highly examplish example:
```rust
pub fn block_from_ip(ip: IpAddr) {
    let rule = firewall::RuleBuilder::default()
        .action(firewall::RuleAction::Block)
        .from(ip)
        .build()
        .unwrap();
    firewall::add_rule(rule);
}
```
But on the other hand the `Foo::builder()` approach works there as well and I don't see any situation where `Foo::builder()` would not work just as well as `FooBuilder::default()`

vision_doc.md
colin-kiegel: Ok, now I get it. You rename `Foo` to `Bar` and the RLS has you covered unless you explicitly named `FooBuilder`. Because that will be renamed implicitly to `BarBuilder` and the RLS will not know about that. :-)

vision_doc.md
colin-kiegel: Actually a bit of both. That's a wall of attributes and it's hard to understand what this is doing. `preserve_attrs(serde)` and all those serde annotations also contribute significantly to that reasoning footprint.

PR #83, opened 2017-04-17
Added macOS support for pre-push tests

TedDriggs:
This incorporates feedback from #72 about running travis tests locally, and adds support for macOS by not trying to use symlinks. It passes on my local box.


colin-kiegel:
Cool, thx. :-)

Issue #43, opened 2017-02-19
Add derives on derived FooBuilder struct


colin-kiegel:
# use case

User wants to have something like `PartialEq` for `FooBuilder`, but can't derive it since `FooBuilder` is already derived by itself.

# possible API

We could add a struct annotation on `Foo`:
- `#[builder(derive(PartialEq, SomethingElse))]`

# example

```rust
#[derive(Builder)]
#[builder(derive(PartialEq, SomethingElse))]
struct Foo {
    foo: Vec<i32>
}
```

Would expand to:
```rust
#[derive(Default, Clone, PartialEq, SomethingElse)]
struct FooBuilder {
    foo: Option<Vec<i32>>
}
```

# discussion

This crate should be conservative and derive as little as possible for `FooBuilder` by default. Because additional derives on `FooBuilder` would add corresponding requirements on the fields of `Foo`.

Status Quo:
- `FooBuilder` impls `Default` and `Clone` (the latter is only needed for the immutable builder pattern)
- All fields on `Foo` already require to implement ...
  - `Clone`, if they have a setter on `FooBuilder`
  - `Default`, if they _don't_ have a setter on `FooBuilder` (.. well, at least with '#15: hidden fields' coming, and this might change with '#3: CustomDefault')

colin-kiegel:
PartialEq on `FooBuilder` might be useful for some unit tests.

TedDriggs:
Enabling a builder to derive serde's Deserialize would be useful for the case of wanting to read an object off the wire (or from a file), applying some overrides, then returning a final constructed object.

colin-kiegel:
Interesting use case.

I see one complication: serde has a lot of field attributes. It would look very ugly if we'd use some pass-through syntax for those field attributes.

However I think that use case can be solved differently. We could derive a second build method on `FooBuilder`, which takes two arguments `fn build_with_defaults(&self, defaults: &Foo) -> Foo` or maybe a mutating variant `fn apply_to(&self, target: &mut Foo)`. Modulo bike shedding. You could then deserialize directly to `Foo` and then apply some incompletely initialized builder to it. Would something like this help you? Or could you describe in more detail, why you need a builder _and_ deserialization at the same time? Where do your overrides come from (what do they depend on)?

PS: I am always happy to hear about what people need and want to achieve. :-)

TedDriggs:
[...]

PR #84, opened 2017-04-17
Added #[builder(derive(...)] for #43

TedDriggs:
Addresses #43; enables builder to have its own additional derive statements. This PR does **not** include attribute forwarding on fields; there are still good questions about the future of that in #79. This does add immediate value for debugging and testing, as it allows generation of equality tests and debug output.

### DIFF
--- a/derive_builder/README.md
+++ b/derive_builder/README.md
@@ -93,6 +93,7 @@ with [cargo-edit](https://github.com/killercup/cargo-edit):
 * **Setter type conversions**: With `#[builder(setter(into))]`, setter methods will be generic over the input types – you can then supply every argument that implements the [`Into`][into] trait for the field type.
 * **Generic structs**: Are also supported, but you **must not** use a type parameter named `VALUE`, if you also activate setter type conversions.
 * **Default values**: You can use `#[builder(default)]` to delegate to the `Default` implementation or any explicit value via `=".."`. This works both on the struct and field level.
+* **Build method suppression**: You can use `#[builder(build_fn(skip))]` to disable auto-implementation of the build method and provide your own.
 *  **no_std support**: Just add `#[builder(no_std)]` to your struct and add `extern crate collections` to your crate. The latter requires the _nightly_ toolchain.
 * **Logging**: If anything works unexpectedly you can enable detailed logs in two steps. First, add `features = ["logging"]` to the `derive_builder` dependency in `Cargo.toml`. Second, set this environment variable before calling cargo `RUST_LOG=derive_builder=trace`.
 
--- a/derive_builder/README.md
+++ b/derive_builder/README.md
@@ -94,7 +94,7 @@ with [cargo-edit](https://github.com/killercup/cargo-edit):
 * **Generic structs**: Are also supported, but you **must not** use a type parameter named `VALUE`, if you also activate setter type conversions.
 * **Default values**: You can use `#[builder(default)]` to delegate to the `Default` implementation or any explicit value via `=".."`. This works both on the struct and field level.
 * **Build method suppression**: You can use `#[builder(build_fn(skip))]` to disable auto-implementation of the build method and provide your own.
-* **Builder derivations**: You can have the builder derive additonal traits using `#[builder(derive(Trait1, Trait2, ...))]`.
+* **Builder derivations**: You can use `#[builder(derive(Trait1, Trait2, ...))]` to have the builder derive additonal traits. All builders derive `Default` and `Clone`, so you should not declare those in this attribute.
 *  **no_std support**: Just add `#[builder(no_std)]` to your struct and add `extern crate collections` to your crate. The latter requires the _nightly_ toolchain.
 * **Logging**: If anything works unexpectedly you can enable detailed logs in two steps. First, add `features = ["logging"]` to the `derive_builder` dependency in `Cargo.toml`. Second, set this environment variable before calling cargo `RUST_LOG=derive_builder=trace`.
 
--- a/derive_builder/README.md
+++ b/derive_builder/README.md
@@ -94,6 +94,7 @@ with [cargo-edit](https://github.com/killercup/cargo-edit):
 * **Generic structs**: Are also supported, but you **must not** use a type parameter named `VALUE`, if you also activate setter type conversions.
 * **Default values**: You can use `#[builder(default)]` to delegate to the `Default` implementation or any explicit value via `=".."`. This works both on the struct and field level.
 * **Build method suppression**: You can use `#[builder(build_fn(skip))]` to disable auto-implementation of the build method and provide your own.
+* **Builder derivations**: You can have the builder derive additonal traits using `#[builder(derive(Trait1, Trait2, ...))]`.
 *  **no_std support**: Just add `#[builder(no_std)]` to your struct and add `extern crate collections` to your crate. The latter requires the _nightly_ toolchain.
 * **Logging**: If anything works unexpectedly you can enable detailed logs in two steps. First, add `features = ["logging"]` to the `derive_builder` dependency in `Cargo.toml`. Second, set this environment variable before calling cargo `RUST_LOG=derive_builder=trace`.
 
--- a/derive_builder/README.md
+++ b/derive_builder/README.md
@@ -91,6 +91,7 @@ with [cargo-edit](https://github.com/killercup/cargo-edit):
 * **Hidden fields**: You can skip setters via `#[builder(setter(skip))]` on each field individually.
 * **Setter visibility**: You can opt into private setter by preceding your struct with `#[builder(private)]`.
 * **Setter type conversions**: With `#[builder(setter(into))]`, setter methods will be generic over the input types – you can then supply every argument that implements the [`Into`][into] trait for the field type.
+* **Builder field visibility**: You can use `#[builder(field(private))]` or `..(public)`, to set field visibility of your builder.
 * **Generic structs**: Are also supported, but you **must not** use a type parameter named `VALUE`, if you also activate setter type conversions.
 * **Default values**: You can use `#[builder(default)]` to delegate to the `Default` implementation or any explicit value via `=".."`. This works both on the struct and field level.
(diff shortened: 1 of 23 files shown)


colin-kiegel:
Cool, will have a look the next days (hopefully soon)! :-)

derive_builder/README.md
```diff
@@ -93,6 +93,7 @@ with [cargo-edit](https://github.com/killercup/cargo-edit):
 * **Setter type conversions**: With `#[builder(setter(into))]`, setter methods will be generic over the input types – you can then supply every argument that implements the [`Into`][into] trait for the field type.
 * **Generic structs**: Are also supported, but you **must not** use a type parameter named `VALUE`, if you also activate setter type conversions.
 * **Default values**: You can use `#[builder(default)]` to delegate to the `Default` implementation or any explicit value via `=".."`. This works both on the struct and field level.
+* **Builder derivations**: You can have the builder derive additonal traits using `#[builder(derive(Trait1, Trait2, ...))]`.
```
colin-kiegel: in `README.md` I prefer example style, e.g.
```You can use `#[builder(derive(Copy, PartialEq, ...))]` to derive additional traits on the builder struct.```

File: derive_builder/tests/derive_trait.rs [hunk 2]
```diff
@@ -0,0 +1,21 @@
+#[macro_use]
+extern crate derive_builder;
+
+#[derive(Debug, Default, Clone)]
+struct NotPartialEq(String);
+
+#[derive(Debug, Clone, Builder)]
+#[builder(derive(Debug, PartialEq, Eq))]
```
colin-kiegel: please add tests that (one of these) is actually implemented.

derive_builder/src/lib.rs
```diff
@@ -367,6 +367,28 @@
...
+//!     foo: u8,
+//!     bar: String,
+//! }
+//! 
+//! fn main() {
+//!    assert_eq!(LoremBuilder::default(), LoremBuilder::default());
+//! }
+//! ```
+//!
+//! Attributes declared for those crates are _not_ forwarded to the fields on the builder.
```
colin-kiegel: What do you mean by _crates_ here? Do you mean struct instead?

Please add another note, that we derive `Default` and `Clone` on the builder unconditionally. Well, unless we change that ^^_Alternatively_ this would be our chance to say `Default` and `Clone` are only derived automatically if the user didn't specify `#[builder(derive)]`. This would allow to opt-out of these derivations - but I feel like they don't do any harm. I lean towards unconditionally deriving them. What do you think?

derive_builder_core/src/builder.rs
```diff
@@ -80,7 +83,7 @@ impl<'a> ToTokens for Builder<'a> {
                    impl_generics);
 
             tokens.append(quote!(
-                #[derive(Default, Clone)]
+                #[derive(Default, Clone #( , #derives)* )]
```
colin-kiegel: I think it would be cleaner, if we move `Default` and `Clone` into the derive options - even, if we set it unconditionally.

colin-kiegel: PS: In that case also remove `Default` and `Clone` from the unit tests - since we already have the testcase `add_derives`.

derive_builder/src/options/struct_mode.rs
```diff
@@ -97,6 +124,7 @@ impl From<OptionsBuilder<StructMode>> for (StructOptions, OptionsBuilder<FieldMo
             builder_visibility: m.builder_vis.unwrap_or(m.build_target_vis),
             builder_pattern: b.builder_pattern.unwrap_or_default(),
             build_target_ident: syn::Ident::new(m.build_target_name),
+            derives: m.derive_traits.unwrap_or_default(),
```
colin-kiegel: Depending on which behaviour we want, I'd suggest either
1. unconditionally push `Debug` and `Clone` onto the derives vector
2. only push `Debug` and `Clone` if they are not yet in the vector
3. or default to ```[`Debug`, `Clone`]```, i.e. `unwrap_or(vec![syn::Ident::new("Debug"), syn::Ident::new("Clone")])`

colin-kiegel: PS: I lean towards 1 or 2.

derive_builder/src/options/struct_mode.rs
```diff
@@ -49,6 +57,25 @@ impl OptionsBuilderMode for StructMode {
...
     }
 
+    fn parse_derive(&mut self, nested: &[syn::NestedMetaItem]) {
+        let mut traits = vec![];
+        for x in nested {
+            match *x {
+                // We don't allow name-value pairs or further nesting here, so
+                // only look for words.
+                syn::NestedMetaItem::MetaItem(syn::MetaItem::Word(ref tr)) => {
+                    traits.push(tr.clone())
```
colin-kiegel: question is, whether we want to do any checks like `tr == "Debug"` | `"Clone"` at *this* place. But I guess it would feel a bit hacky to  have the information about `Debug` and `Clone` being automatically derived scattered about multiple places. It's probably ok, to not have any checks here.

Note: It's worth looking at how the Rust compiler complains about this

# IN
```rust
#[derive(Builder)]
#[builder(derive(Default))]
struct Lorem {
```

# OUT
```
error[E0119]: conflicting implementations of trait `std::default::Default` for type `LoremBuilder`:
 --> derive_builder/tests/derive_trait.rs:7:10
  |
7 | #[derive(Builder)]
  |          -------
  |          |
  |          conflicting implementation for `LoremBuilder`
  |          first implementation here
```

So the error message is ok-ish, but not perfect. I think it would be nicer, if this would either just work without error, or if the builder would panic with a custom message.

derive_builder/src/options/struct_mode.rs
colin-kiegel: If we pick 2, we _could_ emit a (deprecation) warning which informs the user that `Debug` and `Clone` derives are already implicit and that deriving them explicitly is redundant? (deprecations and panics are the only ways we can currently communicate with the user).

derive_builder/src/lib.rs
TedDriggs: That should be "traits", not "crates" - I'll fix that. I agree with unconditional derivation for the `Debug` and `Clone` traits.

derive_builder_core/src/builder.rs
TedDriggs: I thought about this, but as long as these are going to be unconditionally present it felt strange to hide them far from the actual rendering of the builder.

colin-kiegel: TL;DR: Ok,let's just keep it as is :-)

My line of thinking was this:
- `derive_builder_core` is just a toolkit for code generation and should make minimal assumptions
- `derive_builder` populates the settings and sets all defaults - i.e. most decisions should be here

_But_ I just realized that `derive_builder_core` assumes the builder to implement `Clone` for immutable/mutable setters. So we already know in the scope of `derive_builder_core` that we need `Clone` for consistency. With this minimal reasoning rule of thumb the decision about deriving `Clone` should in fact stay in `derive_builder_core`. Now if only had `Default`, we could move at least that to `derive_builder`. But since we have both - let's just keep it as is. :-)

derive_builder/src/options/struct_mode.rs
TedDriggs: I've added a warning at the location in the code review. At the moment it also suppresses forwarding the trait, but I can remove that so we push the warning and then let compilation fail.

derive_builder/tests/derive_trait.rs
TedDriggs: I actually have that; the `defaults` test down below will fail if `LoremBuilder` doesn't implement `PartialEq` and `Debug`. I've made this more explicit with a comment in the test.

colin-kiegel:
Hm, there is again a duplicate commit. 

* [2017-04-17 Added #[builder(derive(...)] `19523fb`
](https://github.com/colin-kiegel/rust-derive-builder/pull/84/commits/19523fb8a6262eaa73ddd415f4fcec92b12fac5c)
* ...
* [2017-04-17 Added #[builder(derive(...)] `e70d8bb`](https://github.com/colin-kiegel/rust-derive-builder/pull/84/commits/e70d8bb45ccd9f7405942feb34474adfbd85eda0)

This makes it harder to review.

According to git log, the commit `19523fb` has a strange parent (`e48b24d`     Colin Kiegel    2017-04-15      clippy). But in github it looks like the parent of `19523fb` was `287877a` Ted Driggs      2017-04-17      Merge remote-tracking branch 'upstream/master' into build_fn.
git_log

I'm no expert with github pull requests, but I would guess that you made a rebase at _some_ point, either directly `git rebase` or maybe indirectly `git pull --rebase`? In that case I would prefer if you did rebases in pull requests only if the changes are so radical, that it's cleaner to start from scratch. If you didn't do a rebase, than I don't understand where this redundant comes from.

derive_builder/tests/derive_trait.rs
```diff
@@ -17,5 +17,7 @@ struct Lorem {
 
 #[test]
 fn defaults() {
+    // This macro requires that the two sides implement `PartialEq` AND `Debug`,
+    // so this one line is testing that the requested traits were in fact generated.
```
colin-kiegel: cool :-)

derive_builder/src/options/struct_mode.rs
```diff
@@ -161,12 +161,19 @@ impl OptionsBuilderMode for StructMode {
...
                 // We don't allow name-value pairs or further nesting here, so
                 // only look for words.
                 syn::NestedMetaItem::MetaItem(syn::MetaItem::Word(ref tr)) => {
-                    traits.push(tr.clone())
+                    match tr.as_ref() {
+                        "Default" | "Clone" => { self.push_deprecation_note(
+                            format!("The `Default` and `Clone` traits are automatically added to all \
+                            builders; explicitly deriving them is unnecessary ({})", where_diag)); 
+                        },
+                        _ => traits.push(tr.clone())
```
colin-kiegel: nice

TedDriggs:
@colin-kiegel you're right, I tried to rebase to clean up the commit history. Thanks for the merge!

Issue #65, opened 2017-04-07
Field-level #[builder(default = __)] easier to use wrong than right


TedDriggs:
```rust
#[derive(Debug, PartialEq, Builder)]
struct Dolor {
    msg: String,
    
    #[builder(default = "self.msg")]
    z_failure: Option<String>
}
```

This produces the following error:

```
error[E0507]: cannot move out of borrowed content
    |
169 |     #[derive(Debug, PartialEq, Builder)]
    |                                ^^^^^^^ cannot move out of borrowed content
```

Alternatively, some really weird things can be accomplished:

```rust
#[derive(Debug, PartialEq, Builder)]
struct Dolor {
    
    #[builder(default)]
    msg: String,

    /// Prints 'hi' if not explicitly initialized.
    #[builder(default = "{println!(\"Hi!\"); Some(\"woot\".to_string())}")]
    z_failure: Option<String>
}
```

The only time the current implementation of this will work at the field level is if the user provides a literal or a value that implements `Copy`. Serde avoids this by requiring the value of `default` to be a function with the signature `Fn() -> T`, which it calls every time it needs a new instance. That feels much more hygienic than the current behavior (and makes using `default` in conjunction with helper or factory functions much cleaner).

@colin-kiegel, how would you feel about combining this with #61 and documenting the breaking change as "aligned `default` behavior with serde"?

colin-kiegel:
Hm, thanks. 

There was a discussion about this [here (comparison with serde / question 1)](https://github.com/colin-kiegel/rust-derive-builder/pull/53#pullrequestreview-28219549) and [in this comment below](https://github.com/colin-kiegel/rust-derive-builder/pull/53#pullrequestreview-28219549):

> I feel like the semantics of `#[serde(default = "some::Function::foo")]` would be too restrictive for derive_builder.
>
> * I suspect a lot of _simple_ cases where delegation to `some::Function::foo` would be overkill and `default="42"` is really all you want
> * For _complex_ cases it would seem beneficial to delegate to _methods_ on `FooBuilder` which can inspect all builder fields.
>
> It would have been nice to share the semantics with a popular crate like serde. But this would have significant drawbacks IMO.

[...]

Issue #70, opened 2017-04-09
skip/rename/private build method


colin-kiegel:
Right now we always generate the build method with a fixed name, and we give it the same privacy as the builder struct.

It would be nice to customize all of that, such that users can wrap a private build method, or just skip it entirely and do their own thing.

This should be rather straightforward to implement. If anyone is interested to start contributing to this crate, this would be a good entry point. :-)

The major question is, how to namespace these options
- `#[builder(build_fn(skip, private, name="finalize"))]`
- `#[builder(buildfn(...))]`
- `#[builder(build(...))]`
- `#[builder(fn(...))]`

All of the possible names I can currently think of look a bit weird.

TedDriggs:
How about `build_fn`?

colin-kiegel:
added that to the list as primary candidate

TedDriggs:
I'm working on this now, if you want to assign the issue to me.

colin-kiegel:
Ok, in the master branch we have now all of the above, except setting the build fn visibility. With the next release I'll probably open another ticket for that, so we can discuss/track it there. :-)

@TedDriggs thank you very much.

colin-kiegel:
released with v0.4.5 - thanks @TedDriggs :-)

PR #85, opened 2017-04-19
#67 Pre-build validation hook

TedDriggs:
This addresses #67 by providing a _pre-construction_ validation hook for structs that use the macro-generated `build` method. This is consistent with @colin-kiegel and @faern's feedback on wanting a place to perform cross-field validation without manually re-implementing the construction step.

At the moment, this does not handle defaults particularly well. I'm working on a sketch of a plan which exposes `get_*`  methods that return one of:

1. `Ok` with the explicit value of that field
1. `Ok` with the default value of that field
1. `Err` with the uninitialized field message

I suspect this will require some rearchitecting of the struct or of the build method to enable populating those prior to calling `validator`, but I'm confident it can be done in a backwards-compatible fashion for anyone who hasn't done really strange things in the `#[builder(default=...)]` expressions (see #65 for examples of things that may cause weirdness).

FYI @colin-kiegel: I've started getting a warning from `cargo test` that the doctests in `lib.rs` aren't being run:

```
WARNING: src/lib.rs -  (line 28) Code block is not currently run as a test, but will in future versions of rustdoc. Please ensure this code block is a runnable test, or use the `ignore` directive.
WARNING: src/lib.rs -  (line 74) Code block is not currently run as a test, but will in future versions of rustdoc. Please ensure this code block is a runnable test, or use the `ignore` directive.
```


faern:
@TedDriggs I thought you said in #67 that you already had a PR for the feature to suppress the build method? Anyway. Would it not be nice to have that as a separate PR since that functionality might be desired and nice to get that merged without being stuck in potential long discussions regarding verification.

derive_builder/examples/validator_fn.rs
```diff
@@ -0,0 +1,37 @@
...
+            }
+        } else {
+            Ok(())
+        }
+    }
+}
+
+fn main() {
+    // If we don't set the field `ipsum`,
+    let x = LoremBuilder::default().build().unwrap();
```
faern: Would it not be nice to set an invalid value and assert the build method returns an `Err`, for demonstrative purposes?

TedDriggs:
@faern I do - it's #70. I think having the validator function attribute be part of the build_fn customization makes clearer that the validator doesn't run if the user chooses not to have the macro generate the `build` method (or whatever the author chooses to name it).

derive_builder/examples/validator_fn.rs
```diff
@@ -0,0 +1,37 @@
+//! This example illustrates the use of `validator` to add a pre-build validation
+//! step.
+
+#[macro_use]
+extern crate derive_builder;
+
+#[derive(Builder, Debug, PartialEq)]
+#[builder(build_fn(validator="LoremBuilder::validate"))]
```
faern: Would it not make sense to limit the value of `validator` to an `Ident` and force it to be a method on the builder? Then it would reduce duplicating the builder name on this line here. Plus since the first argument passed to the validator function would be the builder it would feel strange to have the validator somewhere else.

I prefer adding features as locked down as possible and expand them if there is demand. Like how Rust and the std lib is kind of conservative before adding stuff. If going the other way it becomes hard to keep backwards compatibility since there might be any combination of strange usages in the wild.

faern:
@TedDriggs Yes, having the validator be an argument to `build_fn` might be a very nice idea indeed. But it does not stop us from adding `build_fn` with skip/rename functionality first, separately.

derive_builder/src/options/mod.rs
```diff
@@ -454,7 +459,7 @@ impl<Mode> OptionsBuilder<Mode> where
         trace!("Parsing skip setter `{:?}`", skip);
         self.setter_enabled(!parse_lit_as_bool(skip).unwrap());
     }
-
+
```
faern: Can this be removed again?

TedDriggs: If you want to review just the build function, please have a look at #70. This builds on that, so we can review and merge that one, and I'll update this to sit on top of any changes we make there.

Or did you mean the trailing whitespace?

derive_builder/examples/validator_fn.rs
TedDriggs: I do that in the tests; I've been putting happy paths in the examples.

derive_builder/src/options/struct_mode.rs
```diff
@@ -1,15 +1,18 @@
 use syn;
-use options::{OptionsBuilder, OptionsBuilderMode, parse_lit_as_string, FieldMode, StructOptions};
+use options::{OptionsBuilder, OptionsBuilderMode, parse_lit_as_string, parse_lit_as_bool, parse_lit_as_path, FieldMode, StructOptions};
 use derive_builder_core::{DeprecationNotes, Bindings};
 
 #[derive(Debug, Clone)]
 pub struct StructMode {
+    build_fn_name: Option<String>,
+    build_fn_enabled: bool,
```
faern: I don't know very much about this code base yet, so this is more a question than feedback: Could this not be modeled in only one field? If the `Option<String>` is `None` that means we don't generate a build method. So whoever creates the defaults for this default to `Some("build")` instead of `None`?

derive_builder/src/options/struct_mode.rs
```diff
@@ -24,7 +27,10 @@ impl OptionsBuilder<StructMode> {
             build_target_vis: ast.vis.clone(),
             builder_name: None,
             builder_vis: None,
+            build_fn_enabled: true,
+            build_fn_name: None,
```
faern: Not sure which level of nitpicking is usually applied here. But just want to mention that these should be the first fields to maintain the same order as in the struct def.... :wink:

derive_builder/examples/validator_fn.rs
TedDriggs: It doesn't have to be a static function on the builder. A free function is completely valid, including a generic free function - a generic function with trait bounds gets really compelling. I've used that before in my serialization code, so I'm deliberately aligning this with how [Serde](https://serde.rs/attributes.html#field-attributes) referencing functions in attributes.

You could write `validator="Self::validate"` to avoid duplicating the method name if you wanted.

Changing this to ident would prevent extending it in the future, as there's no way to differentiate `"validate"` the struct-local `Ident` from `"validate"` the in-scope free function.

derive_builder/src/options/struct_mode.rs
TedDriggs: It's probably possible - I'm emulating the existing setter code, which handled these attributes separately.

derive_builder/examples/validator_fn.rs
faern: Fair enough :+1:. I could just not see in what situation it would be usable. If you have a free function with generics and trait bounds, then you would have to implement traits for the builder. I can't say if that is useful, I can't come up with a use case at the moment.

derive_builder/examples/validator_fn.rs
faern: But this `main` does not demonstrate when or where the validator is called, so I would not say it demonstrates the functionality very explicitly. To me it looks more like a example for the `default` feature.

derive_builder/src/options/mod.rs
faern: Yes, the trailing whitespace :)

derive_builder/examples/validator_fn.rs
TedDriggs: Consider a family of builders which have the common requirement that their server IP address must be organization-internal. It would be easy to implement a short trait with one method to expose the server IP and a `validate` function with a default implementation.

derive_builder/src/lib.rs
```diff
@@ -367,6 +367,58 @@
...
+//!             }
+//!         } else {
+//!             Ok(())
+//!         }
+//!     }
+//! }
+//!
+//! fn main() {
+//!     // If we don't set the field `ipsum`,
+//!     let x = LoremBuilder::default().build().unwrap();
```
colin-kiegel: IMO the example should better showcase an invalid input (and not call unwrap, but `assert_eq!(... Err(...))`).

derive_builder/src/lib.rs
```diff
@@ -367,6 +367,58 @@
...
+//! and renaming is done with `#[builder(build_fn(name = "YOUR_NAME"))]`.
+//!
+//! ## Pre-Build Validation
+//! If you're using the provided `build` method, you can declare 
+//! `#[builder(build_fn(validator="path::to::fn"))]` to specify a validator function which gets
+//! access to the builder before construction. 
+//! 
+//! The provided function must have the signature `(&FooBuilder) -> Result<_, String>`; 
+//! the `Ok` variant is not used by the `build` method, and must be accessible from the scope
+//! where the target struct is declared. The path does not need to be fully-qualified, and will
```
colin-kiegel: Please also add a note somewhere, that default values are _not_ validated, i.e. the validation is run before default values are applied. This may not be obvious.

derive_builder/src/options/struct_options.rs
```diff
@@ -5,6 +5,11 @@ use options::DefaultExpression;
 /// These struct options define how the builder is generated.
 #[derive(Debug, Clone)]
 pub struct StructOptions {
+    /// Whether or not this struct should implement its own build method.
+    pub build_fn_enabled: bool,
+
```
colin-kiegel: hm, is this new line intentional?

derive_builder/tests/validator_fn.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+#[test]
+fn incomplete_fields() {
+    let err_msg = LoremBuilder::default().build().unwrap_err();
+    
+    // In this case, the validator should have run but passed because there was no value set for `my_effort`.
+    // Adding private `get` methods which return `Result<T, String>` by pulling in defaults will enable the
+    // validation function to be improved here.
+    assert!(&err_msg != "Don't wear yourself out");
```
colin-kiegel: please try use `assert_ne!(computed, not_expected)` instead - the error message will be more helpful. :-)

derive_builder/tests/validator_fn.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+#[test]
+fn out_of_bounds() {
+    assert_eq!("Don't wear yourself out", &LoremBuilder::default().my_effort(120).build().unwrap_err());
+    assert_eq!("Your rival is cheating", &LoremBuilder::default().rivals_effort(120).build().unwrap_err());
```
colin-kiegel: I prefer to have the field order consistent, i.e. first computed, then expected. That makes it easier to get error messages right. ^^

derive_builder/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - customize setter names via `#[builder(setter(name="..."))]`
+- customize build_fn name via `#[builder(build_fn(name="..."))]`
+- suppress build method generation via `#[builder(build_fn(skip))]`
+- perform pre-build validation via `#[builder(build_fn(validator="path::to::fn"))]`
```
faern: Setter/getter are well known concepts and everyone knows they are methods. If we go with `build_fn` I think `validate_fn` is the name that is most consistent with that and should be used.

colin-kiegel: @faern What do you think about the attribute nesting?
* should it be **nested** like `build_fn(validate_fn)`, e.g. 
  `#[builder(build_fn(name="finish", validate_fn="Self::validate"))]`
* or should it be **flat** like `build_fn, validate_fn`, e.g.
  `#[builder(build_fn(name="finish"), validate_fn="Self::validate")]`

Usually we do nesting, to organize settings. But we already have some kind of exception with `default` which currently only affects the `build_fn` but is not nested. However `default` is also a field specific option and _may_ also affect getters in the future. The reason there was that `#[builder(build_fn(default))]` would have looked a bit weird on a field level. So the situation is definitely a bit different with `validate_fn`.

I think my current favourite would be `validate` + nested, e.g. `#[builder(build_fn(validate="Self::validate"))]`. This way many attributes could be read as imperative verbs, like in _please `validate` this build function via `Self::validate`_, _please `skip` this_, _please `default` this_, _please `name` this `foo`_, ...

It looks like we have a split, where one person each is in favour of `validator`, `validate_fn` or `validate`. ^^

colin-kiegel:
I think we are not very far from reaching consensus here. There are only two controversial topics left:
- bikeshedding how to name the option (`validator`, `validate`, `validate_fn`)
- and the question whether to export and use a `Validate` trait

Note: I'll start a vacation on sunday, but I still see a chance that we might agree on something before that. ;-)

TedDriggs:
If we ship with the path approach, then we can always add trait support once proc-macro crates are allowed to export more things. It would be identical to how we handle `default`:

* In the word form, e.g. `validate`, we try to use the trait
* In the name=value form, e.g. `validate = "Self::validate"` we use the specified path

I'm fine with the name being `validate`; I think that's clear and that name works well for extension to the trait approach in the future.

colin-kiegel:
Ok, cool. That sounds good! :-)

derive_builder/src/lib.rs
TedDriggs: I missed this and a few other comments; I'll get these tomorrow.

colin-kiegel: @Teddriggs I already did the missing comments in  
https://github.com/TedDriggs/rust-derive-builder/commit/bbd830e4f30b9dbd933776185e320ba7734df916. Next time I'll inform you to avoid confusion. :-)

Issue #91, opened 2017-04-25
For generic structs, apply type bounds in impl rather than requiring on base


TedDriggs:
If you run `cargo expand` on this code:

```rust
#[derive(Clone)]
pub struct Foo<T>(T);

fn main() {}
```

You'll see that the generated `Clone` impl has a type bound requiring that `T: ::std::clone::Clone`:

```rust
#![feature(prelude_import)]
#![no_std]
#[prelude_import]
use std::prelude::v1::*;
#[macro_use]
extern crate std as std;
pub struct Foo<T>(T);
#[automatically_derived]
#[allow(unused_qualifications)]
impl<T: ::std::clone::Clone> ::std::clone::Clone for Foo<T> {
    #[inline]
    fn clone(&self) -> Foo<T> {
        match *self {
            Foo(ref __self_0_0) => Foo(::std::clone::Clone::clone(&(*__self_0_0))),
        }
    }
}

fn main() {}
```

This may be possible for the generated builder impl as well: Inherent impl blocks are allowed to have trait bounds beyond what is declared on the base type. Doing so would better align to the standard library and would be slightly more ergonomic.

# Implementation Considerations
1. The `Clone` type bound should not be emitted when using the `owned` builder pattern.
1. The `Default` type bound should not be emitted for generic fields with an explicit default expression.

_Update:_ Turns out that the compiler is perfectly happy with duplicate type bounds:

```rust
// stuff elided...

pub struct Foo<T: ::std::clone::Clone>(T);
#[automatically_derived]
#[allow(unused_qualifications)]
impl<T: ::std::clone::Clone + ::std::clone::Clone> ::std::clone::Clone for Foo<T> {
    #[inline]
    fn clone(&self) -> Foo<T> {
        match *self {
            Foo(ref __self_0_0) => Foo(::std::clone::Clone::clone(&(*__self_0_0))),
        }
    }
}
```

That makes this quite a bit easier.

_Update 2:_ The compiler makes another simplifying assertion:
It requires all generic type params implement `Clone`, rather than all field members' types.

```rust
#[derive(Clone)]
pub struct Lorem<T, U> {
    ipsum: T,
    sit: bool,
    dolor: Doubler<U>,
}

pub struct Doubler<T>(T);

impl Clone for Doubler<SomeUnstableThing> {
    fn clone(&self) -> Self {
        Doubler(SomeUnstableThing::new())
    }
}
```

[...]

PR #92, opened 2017-04-25
Improved type param bound expressions (#91)

TedDriggs:
This addresses #91 for the `Clone` trait using the same principles as the built-in `#[derive(Clone)]` implementation. It's sitting on top of #85, so a diff against that branch may be more useful for seeing the scope of this change. 

@colin-kiegel, I'm not 100% happy with the split between `derive_builder` and `derive_builder_core` at the moment. The `StructOptions` has to pass some of the generics twice or else the generated code won't compile. At the moment that's easy to uphold, but I'm wondering if it would be better to pass `additional_bounds` to `derive_builder_core::Builder` and have it assemble the correct generics internally.

I deliberately didn't put anything about this in `README.md` - I think that the behavior after the PR is more "normal" than the current behavior, so trying to explain it would be confusing.

### DIFF
--- a/derive_builder/src/options/struct_mode.rs
+++ b/derive_builder/src/options/struct_mode.rs
@@ -1,7 +1,7 @@
 use syn;
 use options::{OptionsBuilder, OptionsBuilderMode, parse_lit_as_string, parse_lit_as_bool,
               parse_lit_as_path, FieldMode, StructOptions};
-use derive_builder_core::{BuilderPattern, DeprecationNotes, Bindings};
+use derive_builder_core::{DeprecationNotes, Bindings};
 
 #[derive(Debug, Clone)]
 pub struct StructMode {
@@ -260,7 +260,6 @@ impl From<OptionsBuilder<StructMode>> for (StructOptions, OptionsBuilder<FieldMo
         let bindings = Bindings {
             no_std: b.no_std.unwrap_or(false)
         };
-        let impl_generics = add_generic_bounds(&pattern, &bindings, m.build_target_generics.clone());
 
         let struct_options = StructOptions {
             build_fn_enabled: m.build_fn_enabled.unwrap_or(true),
@@ -276,7 +275,6 @@ impl From<OptionsBuilder<StructMode>> for (StructOptions, OptionsBuilder<FieldMo
             derives: m.derive_traits.unwrap_or_default(),
             deprecation_notes: m.deprecation_notes,
             generics: m.build_target_generics,
-            inherent_generics: impl_generics,
             struct_size_hint: m.struct_size_hint,
             bindings: bindings,
             default_expression: struct_default_expression,
@@ -285,26 +283,4 @@ impl From<OptionsBuilder<StructMode>> for (StructOptions, OptionsBuilder<FieldMo
 
         (struct_options, field_defaults)
     }
-}
-
-#[allow(dead_code)]
-fn add_generic_bounds(pattern: &BuilderPattern,
-                      bindings: &Bindings,
-                      mut generics: syn::Generics)
-                      -> syn::Generics {
-    if !pattern.requires_clone() {
-        return generics;
-    }
-
-    for mut typ in generics.ty_params.iter_mut() {
-        typ.bounds.push(syn::TyParamBound::Trait(
-            syn::PolyTraitRef {
-                trait_ref: syn::parse_path(bindings.clone_trait().as_str()).unwrap(),
-                bound_lifetimes: vec![],
-            },
-            syn::TraitBoundModifier::None
-        ))
-    }
-
-    generics
 }
--- a/derive_builder/src/options/struct_mode.rs
+++ b/derive_builder/src/options/struct_mode.rs
@@ -1,6 +1,7 @@
 use syn;
-use options::{OptionsBuilder, OptionsBuilderMode, parse_lit_as_string, parse_lit_as_bool, parse_lit_as_path, FieldMode, StructOptions};
-use derive_builder_core::{DeprecationNotes, Bindings};
+use options::{OptionsBuilder, OptionsBuilderMode, parse_lit_as_string, parse_lit_as_bool,
+              parse_lit_as_path, FieldMode, StructOptions};
+use derive_builder_core::{BuilderPattern, DeprecationNotes, Bindings};
 
 #[derive(Debug, Clone)]
 pub struct StructMode {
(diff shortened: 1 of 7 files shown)


derive_builder/src/options/struct_mode.rs
```diff
@@ -233,7 +230,7 @@ impl From<OptionsBuilder<StructMode>> for (StructOptions, OptionsBuilder<FieldMo
                 `field(<vis>)` to the builder attribute on the struct. (Found {})",
                 where_diagnostics));
         }
-        
+
```
TedDriggs: My editor keeps insisting on doing this, and I'm not sure why. I'll see if I can find a setting that will make it stop.

derive_builder/tests/bounds_generation.rs
```diff
@@ -0,0 +1,55 @@
...
+}
+
+#[test]
+fn u16_lorems() {
+    assert_eq!(make_u16_lorem(10), build_u16_lorem(10));
+}
+
+#[test]
+fn dolor_lorems() {
+    assert_eq!(make_default_lorem(), make_dolor_lorem(Dolor::default()));
```
colin-kiegel: Hm, in a test summary `dolor_lorems` will not be very informative. :-)

Please try to describe the test property in the function name.

If I understand this correctly, it's actually a _compile-test_ in the sense that _hey this compiles now, but it didn't before_. This test is actually only asserting something (almost) arbitrary in order to show up in the test summary, right?

In that case I would prefer to turn it into something like this

```rust
#[test]
fn generic_field_without_clone_compiles() {
    // This is actually a compile test, if we can _name_ `LoremBuilder<Dolor>`, we are already fine.
    // But we will not be able to call any setters or build method on it, because that would
    // require the field to implement `Clone`.
    let _: LoremBuilder<Dolor>;
}
```

derive_builder/tests/bounds_generation.rs
```diff
@@ -0,0 +1,55 @@
...
+fn build_u16_lorem(v: u16) -> Lorem<u16> {
+    LoremBuilder::default()
+        .ipsum(v)
+        .build()
+        .unwrap()
+}
+
+#[test]
+fn u16_lorems() {
+    assert_eq!(make_u16_lorem(10), build_u16_lorem(10));
```
colin-kiegel: please use a more informative title (see below). Also since most methods are used only once, I think it's easier to read inlined, no?

E.g. 
```rust
#[test]
fn generic_field_with_clone_has_builder_impl() {
    assert_eq!(LoremBuilder::default()
                   .ipsum(10)
                   .build()
                   .unwrap(),
               Lorem {
                   ipsum: 10
               });
}
```

derive_builder/tests/bounds_generation.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[test]
+fn dolor_lorems() {
+    assert_eq!(make_default_lorem(), make_dolor_lorem(Dolor::default()));
+}
+
+#[test]
+fn type_inference() {
+    assert_eq!(LoremBuilder::<String>::default().ipsum("".to_string()).build().unwrap(), 
+               make_default_lorem());
```
colin-kiegel: title is fine, but please put the thing you want to test on the left side and try to keep the other side simple. E.g. something like this perhaps?

```rust
#[test]
fn type_inference() {
    assert_eq!(make_default_lorem(),
               Lorem { ipsum: "".to_string() });
}
```

I prefer a pattern where the _left side does all fragile computations and the right side is as robust/straight-forward as possible_. If there is an error this will make it easier to trace it down, because you can concentrate on one arm of the equality. :-)

derive_builder/tests/bounds_generation.rs
```diff
@@ -0,0 +1,55 @@
...
+extern crate derive_builder;
+
+/// A struct that deliberately doesn't implement `Clone`.
+#[derive(Debug, Default, PartialEq, Eq)]
+pub struct Dolor(String);
+
+/// Notice that this type derives Builder without disallowing
+/// `Lorem<Dolor>`.
+#[derive(Debug, Clone, Builder, PartialEq, Eq)]
+#[builder(field(private), setter(into))]
```
colin-kiegel: please remove the attributes you don't need. Both `field(private)` and `setter(into)` are not necessary IMO - right?

derive_builder/tests/bounds_generation.rs
colin-kiegel: I think I'd also like another test - like this:
```rust
#[test]
fn generic_field_without_clone_has_owned_builder() {
    assert_eq!(OwnedLoremBuilder::default()
                   .ipsum(Dolor::default())
                   .build()
                   .unwrap(),
               OwnedLorem {
                   ipsum: Dolor::default()
               });
}
```

colin-kiegel:
PS: Yeah, I agree that `README` is only to showcase important stuff. But this sentence is not true anymore: 
> `README.md:99:127` All builders derive `Default` and `Clone`, so you should not declare those in this attribute.

In `lib.rs:491` this can be removed:
> //! - Generic structs need the boundary `where T: std::clone::Clone` if
> //!   used in combination with the immutable/mutable pattern

derive_builder/tests/bounds_generation.rs
TedDriggs: `field(private)` was needed to squelch the warnings about public fields. You're right that I don't need `setter(into)`; I'll remove that.

TedDriggs:
I'm partway through moving this logic over; turns out it also requires the `derive_builder_core::Builder` struct to know about the bindings. Not a big deal, but wanted to flag it for you.

`README.md:99:127` is true; the builder derives `Default` and `Clone`, so you shouldn't declare those in the `derive(...)` attribute or we'll emit a warning. The built-in derives for those types already apply bounds checks like the ones I'm adding for the builder's inherent impl.

I'll remove the line in `lib.rs`.

derive_builder_core/src/builder.rs
```diff
@@ -169,24 +174,40 @@ mod tests {
             }
         ));
     }
+    
+    fn clone_bound(mut generics: syn::Generics) -> syn::Generics {
+        for mut typ in generics.ty_params.iter_mut() {
+            typ.bounds.push(syn::TyParamBound::Trait(
+                syn::PolyTraitRef {
+                    trait_ref: syn::parse_path("::std::clone::Clone").unwrap(),
```
TedDriggs: Done.

derive_builder/tests/bounds_generation.rs
TedDriggs: Tests have been cleaned up and expanded. As far as your second comment, you got a test that looks _exactly_ like that 😛

derive_builder/tests/bounds_generation.rs
colin-kiegel: I know, we have deprecations all over the place. :-)

But we also run all tests a second time with `feature = "private_fields"`. So I think it's best not to add `#[builder(fields(private))]` on every test, but rather live with the deprecations for one cycle.

TedDriggs: I ended up keeping it, because I added a test that uses it.

derive_builder_core/src/builder.rs
```diff
@@ -128,6 +134,36 @@ impl<'a> Builder<'a> {
...
+
+            for mut typ in generics.ty_params.iter_mut() {
+                typ.bounds.push(clone_bound.clone());
+            }
+
+            generics
+        } else {
+            Default::default()
+        }
+    }
```
colin-kiegel: very nice! :+1:

TedDriggs: Thanks!

PR #94, opened 2017-04-26
refactoring: re-export custom derive with additional items (WIP)

colin-kiegel:
This allows to add other item definitions as described in https://users.rust-lang.org/t/10546/2
* renamed `derive_builder` -> `derive_builder_macro`
* added new `derive_builder`, which re-exports `derive_builder_macro`
* moved all testcases into new crate `testsuite`
* moved `no_std` testcase into a separate crate inside `testsuite`.
  This way we have better control, e.g.
  - panic="abort" compiler flag
  - only run on nightly
  - compile `derive_builder` with no_std feature

**WIP**
- The [`no_std` test is still failing](https://travis-ci.org/colin-kiegel/rust-derive-builder/jobs/226205835#L420-L438). Somehow `derive_builder` still re-exports `std`. This is probably due to the glob-export in `derive_builder/src/lib.rs:530`

  ```rust
  pub use derive_builder_macro::*;
  ```

  But it would seem very difficult to write `derive_builder_macro` without `std`. There must be a different way - especially since serde seems to be doing something like this successfully. [Could this be the trick?](https://github.com/serde-rs/serde/blob/6fbf40b83c65ba2bfe9e94276482b13dd979ac63/serde/src/lib.rs#L116-L125)

cc @TedDriggs


TedDriggs:
I'd request we merge #92 before embarking on this, or else I'll have a hell of a time merging that later. I can look at this next week and see if I can figure out what's causing problems.

colin-kiegel:
Done. I also updated this branch. :-)

It was surprisingly simple - git was able to figure everything out - except of the location of a new file.
```bash
git checkout reexport_macro_with_items
git merge master
git mv derive_builder/tests/bounds_generation.rs testsuite/tests/
git commit --amend
```

TedDriggs:
I think I've got most of it figured out now; `dev/githook.sh` passes cleanly, which is promising. I've added the `pub mod export`, and as a result gutted most of the `Bindings` struct (which is now an enum with `Std` and `NoStd`).

colin-kiegel:
@TedDriggs Ok, cool. I merged your changes into from your fork into this branch.

`dev/nightlytests.sh` seems to be still failing (at least for me) - I get the same error as before
```
error[E0152]: duplicate lang item found: `eh_personality`.
  --> no_std/tests/no_std.rs:51:1
   |
51 | pub extern  fn eh_personality() {}
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: first defined in crate `panic_unwind`.


error[E0152]: duplicate lang item found: `panic_fmt`.
  --> no_std/tests/no_std.rs:61:1
   |
61 | / pub extern fn rust_begin_panic(_msg: core::fmt::Arguments,
62 | |                                _file: &'static str,
63 | |                                _line: u32) -> ! {
64 | |     unsafe { intrinsics::abort() }
65 | | }
   | |_^
   |
   = note: first defined in crate `std`.

```

**Note**: You should also be able to push directly to the branch in _this_ repository. If your local branch is also called `reexport_macro_with_items`, then you should be able to do something like:
```bash
git remote add upstream git@github.com:colin-kiegel/rust-derive-builder.git
git push --set-upstream upstream reexport_macro_with_items
```

derive_builder/src/lib.rs
```diff
@@ -519,15 +519,21 @@
...
 
 #[doc(hidden)]
 pub mod export {
-
+    pub use ::core::clone::Clone;
+    pub use ::core::convert::Into;
+    pub use ::core::default::Default;
+    pub use ::core::marker::PhantomData;
+    pub use ::core::option::Option::{self, Some, None};
+    pub use ::core::result::Result::{self, Ok, Err};
```
colin-kiegel: I wonder if this can lead to problems if crate A exports `FooBuilder` and crate B wants to use that builder. Without depending on `derive_builder`, all those re-exports would not be in scope of crate B.

We should double check if that is a problem.

(Serde seems to do somethings similar https://github.com/serde-rs/serde/blob/6fbf40b83c65ba2bfe9e94276482b13dd979ac63/serde/src/export.rs)

derive_builder_macro/src/options/field_mode.rs
```diff
@@ -173,16 +172,18 @@ fn filter_attr(attr: &&syn::Attribute) -> bool {
 
     if attr.is_sugared_doc == true {
         if let syn::MetaItem::NameValue(ref ident, _) = attr.value {
-            // example:
-            // Attribute { style: Outer, value: NameValue(Ident("doc"), Str("/// This is a doc comment for a field", Cooked)), is_sugared_doc: true }
+    // example:
+    // Attribute { style: Outer, value: NameValue(Ident("doc"), Str("/// This is a
+    // doc comment for a field", Cooked)), is_sugared_doc: true }
```
colin-kiegel: the indentation should not change here

derive_builder_macro/src/options/field_mode.rs
```diff
@@ -173,16 +172,18 @@ fn filter_attr(attr: &&syn::Attribute) -> bool {
...
                 return true
             }
         }
     } else {
         if let syn::MetaItem::List(ref ident, _) = attr.value {
-            // example:
-            // Attribute { style: Outer, value: List(Ident("allow"), [MetaItem(Word(Ident("non_snake_case")))]), is_sugared_doc: false }
+    // example:
+    // Attribute { style: Outer, value: List(Ident("allow"),
+    // [MetaItem(Word(Ident("non_snake_case")))]), is_sugared_doc: false }
```
colin-kiegel: the indentation should not change here

testsuite/tests/no_std
```diff
@@ -1 +0,0 @@
-../no_std/tests/
```
colin-kiegel: didn't you like this symlink?

I thought it would make it easier to find the `no_std` testcases.

TedDriggs: It was causing build problems for me for some reason.

derive_builder/src/lib.rs
TedDriggs: It doesn't seem to; it looks like the system is smart enough to know that these are actually the same ones as in `::std`, or in `::core`

TedDriggs:
@colin-kiegel as a heads-up, I'm going to take another stab at this. Since we don't support generating multiple builders off a single struct, it still makes sense to me to expose a `Builder` trait with an associated type for the kind of builder returned. This would give us a means to expose a structured "Result" type.

One major change though: Somehow the crate name `builder` is still available. Rather than massively disrupt this crate, I'm going to make a lightweight `builder` crate that in turn depends on `derive_builder` for now.

Issue #98, opened 2017-05-08
Skipped fields get wrong default values


TedDriggs:
Found this during refactoring:

```rust
#[macro_use]
extern crate derive_builder;

#[derive(Debug, PartialEq, Eq, Builder)]
#[builder(default)]
struct Lorem {
    #[builder(setter(skip))]
    foo: u8,
    bar: String,
}

impl Default for Lorem {
    fn default() -> Self {
        Lorem {
            foo: 10,
            bar: Default::default(),
        }
    }
}

#[test]
fn skipped_field_gets_struct_default() {
    assert_eq!(LoremBuilder::default().bar("hello".to_string()).build().unwrap(), Lorem {
        foo: 10,
        bar: "hello".to_string(),
    });
}
```

Output:
```
running 1 test
test skipped_field_gets_struct_default ... FAILED

failures:

---- skipped_field_gets_struct_default stdout ----
        thread 'skipped_field_gets_struct_default' panicked at 'assertion failed: `(left == right)` (left: `Lorem { foo: 0, bar: "hello" }`, right: `Lorem { foo: 10, bar: "hello" }`)', tests/skip_defaults.rs:23
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

@colin-kiegel, I can submit a PR with the new test, but at the moment I don't have a fix so it won't pass.

colin-kiegel:
Hm, nice catch. :-)

I think the test would be a good starting point for a fix. But I'll wait with merging, to avoid the impression that we have some kind of regression, ok?

colin-kiegel:
@TedDriggs can this be closed?

TedDriggs:
Yes, I don't think this has repro'd since then.

PR #101, opened 2017-05-30
Add test demonstrating #98 works

TedDriggs:
This supersedes previous test for #98, and shows that something has fixed it.


colin-kiegel:
Maybe it's because of `#[cfg(feature = "struct_default")]`? I didn't see this line in #98. Anyway good to see that this is now a non-issue. :-)

PR #118, opened 2018-05-16
Update gitignore to exclude files generated by RLS

TedDriggs:
Opened the project today and was blasted by untracked files created by RLS. Not sure if that's intentional or a bug on their side, but in the interest of unblocking development I've ignored it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/rls
 target
 Cargo.lock
 .*


Dylan-DPC:
Thanks.

badboy:
That ... should not happen and sounds more like a bug with rls? Did you report that upstream?

nrc:
It's a known bug. Unfortunately it needs a point release of stable to fix, which I don't think will happen until next week. In the meantime you can use beta or nightly branch for a working RLS. Sorry!

PR #111, opened 2018-03-05
Disable derive clone on owned pattern

azriel91:
Attempts to address #97, here's the quick summary:

* Made `Clone` optionally derived on the `Builder`, depending on the `BuilderPattern` on the struct as well as on each of the fields. I'm not sure if the way it calculates and passes the "mark this as should derive `Clone`" is a good way &mdash; it was the quickest way to get *most* of the tests to pass.
* Ran current Rustfmt over the repository.
* Updated top level docs (minor).
* Updated changelog.

The `./dev/checkfeatures.sh` run fails, but I couldn't figure out how to make it pass:

* `getopts 0.2.17` fails to compile:

    ```
    $ ./dev/checkfeatures.sh

    Running `cd derive_builder && rustup run 1.15.0 cargo test --all --color always --features "skeptic_tests"`
       Compiling getopts v0.2.17
    error[E0301]: cannot mutably borrow in a pattern guard
       --> /home/azriel/.cargo/registry/src/github.com-1ecc6299db9ec823/getopts-0.2.17/src/lib.rs:405:73
        |
    405 |                         } else if was_long || name_pos < names.len() || args.peek().map_or(true, |n| is_arg(&n)) {
        |                                                                         ^^^^ borrowed mutably in pattern guard

    error: aborting due to previous error

    error: Could not compile `getopts`.
    ```

* `Cargo.lock` says `compiletest_rs 0.3.7` depends on `getopts 0.2.17`.
* `Cargo.toml` in `derive_builder` says:

    ```toml
    [dependencies]
    # ... omitted
    compiletest_rs = { version = "0.3.3", optional = true }
    ```

Which, to me says "use version `0.3.3` of compiletest_rs" but cargo is disobeying that. Then I became sad because I tried munging dependencies, but cargo always chose `0.3.7` of `compiletest_rs`. I did delete `Cargo.lock` and try again + some other variations. Maybe this issue would solve that: https://github.com/rust-lang/cargo/issues/4100

<details>
<summary><i>Re-enactment of how I did this PR</I></summary>

* Ripgrep: `rg -F owned`.
* Hack code.
* `cargo test`
* `rg -F '::Owned'`.
* Fix compile errors.
* Read `CONTRIBUTING.md`
* Do what it says, ish 😀.
</details>

PR #122, opened 2018-05-30
Only derive Clone on builders when necessary

### LINKED ISSUE #97: Owned builders should not #[derive(Clone)]
Hi,

my struct has a non-clonable field, thus i want to disable the derived impl of *Clone* trait for the generated builder.

TedDriggs:
Builders using the owned pattern for all fields don't need Clone derived for them,
and therefore can be used with generics that don't support cloning.

Fixes #97, and obviates #111 

@azriel91, please review this.

### DIFF
--- a/derive_builder/src/options/darling_opts.rs
+++ b/derive_builder/src/options/darling_opts.rs
@@ -319,14 +319,10 @@ impl Options {
             .fields
     }
 
-    /// A builder requires `Clone` to be derived if any of its fields use the mutable or immutable pattern.
-    ///
-    /// This can technically be expressed two ways:
-    ///
-    /// 1. The caller chooses the `owned` pattern at the struct level
-    /// 1. The caller chooses the `owned` pattern on every field
+    /// A builder requires `Clone` to be derived if its build method or any of its setters
+    /// use the mutable or immutable pattern.
     pub fn requires_clone(&self) -> bool {
-        self.fields().any(|f| f.pattern().requires_clone())
+        self.pattern.requires_clone() || self.fields().any(|f| f.pattern().requires_clone())
     }
 
     /// Get an iterator over the input struct's fields which pulls fallback
--- a/derive_builder/src/options/darling_opts.rs
+++ b/derive_builder/src/options/darling_opts.rs
@@ -294,6 +294,16 @@ impl Options {
             .fields
     }
 
+    /// A builder requires `Clone` to be derived if any of its fields use the mutable or immutable pattern.
+    ///
+    /// This can technically be expressed two ways:
+    ///
+    /// 1. The caller chooses the `owned` pattern at the struct level
+    /// 1. The caller chooses the `owned` pattern on every field
+    pub fn requires_clone(&self) -> bool {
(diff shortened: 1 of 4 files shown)


azriel91:
I'll get to this in the upcoming Tuesday, going to be temporarily inactive till then

derive_builder_core/src/builder.rs
```diff
@@ -84,6 +89,24 @@ impl<'a> ToTokens for Builder<'a> {
...
+
+            // Create the comma-separated set of derived traits for the builder
+            let derived_traits = {
+                let default_trait: Ident = parse_quote!(Default);
+                let clone_trait: Ident = parse_quote!(Clone);
+
+                let mut traits: Punctuated<&Ident, Token![,]> = Default::default();
+                traits.push(&default_trait);
+
+                if self.must_derive_clone {
```
azriel91: Should this also check if the builder pattern is non-owned?
The tests use the `default_builder!` macro, which has `must_derive_clone = true`, but `Options#as_builder` only checks if any fields require clone. I've tried tracing the code but haven't found the answer

derive_builder/src/options/darling_opts.rs
```diff
@@ -319,6 +319,16 @@ impl Options {
             .fields
     }
 
+    /// A builder requires `Clone` to be derived if any of its fields use the mutable or immutable pattern.
+    ///
+    /// This can technically be expressed two ways:
```
azriel91: Feels weird to explain how to express that `Clone` is required, and it should be the non-owned patterns that cause `Clone` to be required. I think the first sentence by itself is enough

derive_builder_core/src/builder.rs
TedDriggs: This shouldn't, but I was wondering if the function in `darling_opts` should.

derive_builder/src/options/darling_opts.rs
TedDriggs: You're right; I'll remove the second part of this comment.

derive_builder/src/options/darling_opts.rs
```diff
@@ -319,6 +319,16 @@ impl Options {
...
+    ///
+    /// 1. The caller chooses the `owned` pattern at the struct level
+    /// 1. The caller chooses the `owned` pattern on every field
+    pub fn requires_clone(&self) -> bool {
+        self.fields().any(|f| f.pattern().requires_clone())
```
TedDriggs: @azriel91 if there's a place where we should check the struct's pattern, it'd be here. However, I don't think we strictly _need_ to; unless I'm missing something, it only matters what the individual fields are doing: A mutable zero-field builder shouldn't need `Clone` to operate.

azriel91: True that (though one would wonder why you'd have a zero field builder :stuck_out_tongue:)

TedDriggs: Wait, I forgot that `build` takes a `&self` if the struct-level one isn't marked `owned`. I've updated the logic accordingly.

PR #126, opened 2018-05-30
Bump rust min version in skeptic and README

### LINKED ISSUE #123: dev/checkfeatures.sh should not rely on rust v1.15
We've moved to a minimum rust version of 1.18.0.

@azriel91, do you mind taking your changes from #111 and making a new PR with them for this issue?

TedDriggs:
@azriel91, please take a look at this if you get a chance. I think your version was more expansive and better.

Fixes #123

### DIFF
--- a/derive_builder/README.md
+++ b/derive_builder/README.md
@@ -1,5 +1,5 @@
 [![Build status](https://travis-ci.org/colin-kiegel/rust-derive-builder.svg?branch=master)](https://travis-ci.org/colin-kiegel/rust-derive-builder)
-[![Rust version]( https://img.shields.io/badge/rust-1.15+-blue.svg)]()
+[![Rust version]( https://img.shields.io/badge/rust-1.18+-blue.svg)]()
 [![Documentation](https://docs.rs/derive_builder/badge.svg)](https://docs.rs/derive_builder)
 [![Latest version](https://img.shields.io/crates/v/derive_builder.svg)](https://crates.io/crates/derive_builder)
 [![All downloads](https://img.shields.io/crates/d/derive_builder.svg)](https://crates.io/crates/derive_builder)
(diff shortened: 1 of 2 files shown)


Dylan-DPC:
I prefer taking a slight bigger jump and targetting 1.22.0 as minimum version instead (allows us to use features that were introduced in 1.22.0 and also ensures we make smaller jumps in future)

TedDriggs:
> use features that were introduced in 1.22.0

Which features did you have in mind?

Dylan-DPC:
Right now I don't see anything. But what commonly happens is that if someone wants to PR a change with features that are introduced in the later stages, you end up introducing feature gates ot sticking to older features.

TedDriggs:
> you end up introducing feature gates ot sticking to older features.

My preference at that time would be to bump the min version and the crate's minor version. The [2017 Rust survey](https://blog.rust-lang.org/2017/09/05/Rust-2017-Survey-Results.html#path-to-stability) says most people are on latest stable, so I'm not too worried about leaving people behind when we do have to make those jumps.

Dylan-DPC:
I agree. But then say if there is a crate that is in post-stable stage (post 1.0 release) and every time a dependency shifts its minimum version and they have to do so accordingly, they are in a tough spot. They can't just shift their minimum rust version, because it is a breaking change and they have to release a new major version and wouldn't to do that every time a dependent crate shifts its minimum rust version.

TedDriggs:
> But then say if there is a crate that is in post-stable stage (post 1.0 release)

I don't think this crate is anywhere near stabilization, so I'd rather cross that bridge if we come to it. `darling` is a mandatory dependency in 0.6, and it has a hard dependency on Rust 1.18. I'd rather we not set the minimum version higher than necessary.

Dylan-DPC:
> I don't think this crate is anywhere near stabilization, so I'd rather cross that bridge if we come to it. 

Wasn't referring to this crate, but any crate that uses this one as a dependency. The general consensus on minimum rust version is (stable - 2). Anyway, i'm fine with 1.18 but just that we shouldn't be jumping versions too often.

PR #129, opened 2018-05-30
Handle generics in build method struct default instantiation

### LINKED ISSUE #127: Struct-level defaults should work with structs that have generics
On master, attempt to use `builder(default)` on a struct that has a generic parameter. This produces code that fails to compile, complaining that it expected one type parameter but got 0.

```rust
    pub fn build(&self) -> ::std::result::Result<RenderItem<T>, ::std::string::String> {
        let __default: RenderItem = { ::std::default::Default::default() };
    }
```

The issue appears to be here: The type of `__default` needs the generic type parameters but it doesn't get them.

TedDriggs:
Fixes #127 for 0.5.2 without requiring the full update to 0.6.0

### DIFF
--- a/derive_builder/CHANGELOG.md
+++ b/derive_builder/CHANGELOG.md
@@ -2,6 +2,11 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [0.5.2] - Unreleased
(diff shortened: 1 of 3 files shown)


TedDriggs:
@colin-kiegel do I have permissions to publish a release?

PR #130, opened 2018-06-04
Specify build_fn(private) in builder attributes

### LINKED ISSUE #89: build function visibility
Since PR #70 added skipping and renaming of the build function. But we can't set the visibility. It might be useful to implement hidden build functions to be able to wrap them in a custom public build function.

This should be rather straightforward to implement. If anyone is interested to start contributing to this crate, this would be a good entry point. :-)

- `#[builder(build_fn(private))]` (cmp with `#[builder(setter(private))]`)

TedDriggs:
Fixes #89


TedDriggs:
This failure isn't the result of this change; instead, it's due to nightly breaking thanks to #131

PR #143, opened 2019-03-13
chore: Update pretty_assertions to 0.6

ignatenkobrain:
Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>


TedDriggs:
Unfortunately, it looks like the new version of pretty-assertions isn't compatible with 1.18, which this library needs to support at least until 0.7.

Issue #116, opened 2018-05-03
Feature: Allow "strip option" on setters


TedDriggs:
Consider a case where I have the following:

```rust
// The `from_variants` crate will generate From<String> for Foo and From<u32> for Foo impls
#[derive(FromVariants)]
pub enum Foo {
    Bar(String),
    Baz(u32),
}

#[derive(Builder)]
#[builder(setter(into))]
pub struct Built {
    value: Option<Foo>,
    other: Option<Foo>,
}
```

I'd like the user of this to write:

```rust
fn make_built() -> Built {
    BuiltBuilder::default()
        .value(17)
        .build()
        .unwrap()
}
```

In this case, `value` should be initialized to `Some(Foo::Baz(17))` and `other` should remain `None`.

What I'm imagining is there's some new setter option, called `always_some` or `pierce_option` or `strip_option`. It would work at the struct and field level, though using it on a field which wasn't an option would produce an error. When it's present, if the type is exactly "Option" then we grab the inner type and use that for the setter, wrapping the provided value in a `Some` in the setter.

# Problems
1. We have to break open the type to figure out the non-optional part, which isn't always going to work.

davidB:
Any suggestion about how to start to try to implement this feature ? (I never made a derive, but I'll provide a PR if I have something usefull).

Context, I would like to provide an easy way to create struct generated automaticaly from a json-schema (so lot Option, and enum,...).

TedDriggs:
Thanks for offering to look into it. For this feature, you're going to need two parts:

1. An update to the `Field` struct in `darling_opt.rs` [here](https://github.com/colin-kiegel/rust-derive-builder/blob/master/derive_builder/src/options/darling_opts.rs#L177) where you allow the user to pass the optional `strip_option` flag.
2. An update to the generated code which attempts to look inside the type of the field and get the underlying value if it's an `Option<T>`, or emit a compile error otherwise.

davidB:
@TedDriggs 
I guess this ticket could be closed.
When do you plan to "cargo publish 0.7.2" (I need it to publish my own crate ;-) ) ?

TedDriggs:
Sorry, running the job to publish now.

PR #145, opened 2019-05-21
add strip_option to setter

davidB:
related to ticket #116 but with some difference against original request (mark with (!)):
- `strip_option` is a flag of setter
-  it can be applied at field or struct level
- (!) unset value is set to None, only if the field is builder is flagged with `default`
- (!) if the flag is applied on non Option<T>, then it is ignored (as it can be set on a struct with non Option field without generate error, less boiler plate to use IMHO, and match my target usage)

see change in lib.rs and test for usage.

### DIFF
--- a/derive_builder_core/src/setter.rs
+++ b/derive_builder_core/src/setter.rs
@@ -71,7 +71,7 @@ impl<'a> ToTokens for Setter<'a> {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         if self.enabled {
             trace!("Deriving setter for `{}`.", self.field_ident);
-            let ty = self.field_type;
+            let field_type = self.field_type;
             let pattern = self.pattern;
             let vis = &self.visibility;
             let field_ident = self.field_ident;
@@ -81,6 +81,13 @@ impl<'a> ToTokens for Setter<'a> {
             let clone = self.bindings.clone_trait();
             let option = self.bindings.option_ty();
             let into = self.bindings.into_trait();
+            let (ty, stripped_option) = match self.strip_option {
+                false => (field_type, false),
+                true => match extract_type_from_option(field_type) {
+                    None => (field_type, false),
+                    Some(ty) => (ty, true),
+                },
+            };
 
             let self_param: TokenStream;
             let return_ty: TokenStream;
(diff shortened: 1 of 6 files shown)


derive_builder_core/src/setter.rs
```diff
@@ -152,6 +164,46 @@ impl<'a> ToTokens for Setter<'a> {
     }
 }
 
+// adapted from https://stackoverflow.com/a/55277337/469066
+// Note that since syn is a parser, it works with tokens.
+// We cannot know for sure that this is an Option. The user could, for example, type
+// - `std::option::Option`
+// - `type MaybeString = std::option::Option<String>`
+// We cannot handle those arbitrary names.
```
TedDriggs: I agree that we can't handle arbitrary aliases, but is there a reason we can't handle `std::option::Option`, `::std::option::Option` (and the same for `core`)? I think that would improve our coverage to almost 100% (because who redefines/aliases `Option`?)

davidB: I though the same, I'll add string comparison for those cases.

davidB: it's more complicated than string comparison, it will take more time than expected.

derive_builder_core/src/setter.rs
```diff
@@ -166,42 +166,56 @@ impl<'a> ToTokens for Setter<'a> {
...
-                        _ => None,
-                    }
-                })
-                .and_then(|generic_arg| match *generic_arg.into_value() {
-                    GenericArgument::Type(ref ty) => Some(ty),
-                    _ => None,
-                })
-        }
-        _ => None,
+    // TODO store (with lazy static) precomputed parsing of Option when support of rust 1.18 will be removed (incompatible with lazy_static)
```
TedDriggs: I'm fine merging as-is, but do you know if there's an older version of `lazy_static` we could use that is compatible with `rustc 1.18`? Unfortunately I don't have a good way of knowing the versions of Rust where this crate is being used, so I don't have a timetable for dropping this support.

davidB: I tried without success the latest version of lazy_static and the version 1.0.0. using lazy_static was a better optimization  for a previous implementation I made.
TBH, the support 1.18 generate some trouble, I develop.

Issue #110, opened 2018-03-03
Emit field - but not setter - to support setter defined in user code


lobisquit:
I would like to use, in my code, the useful `#[builder(setter(skip))]` option to write a custom setter for a struct field.

Expanding macros I noticed though that such field is marked in the builder as `std::marker::PhantomData<...>`, and I don't know how to deal with it.
Other generated setters assign actual value to common `Option<...>`: how can the same be done with `PhantomData<...>`?

fluffysquirrels:
A PhantomData instance cannot be written to, it's just there for the
compiler to track types. Can you try writing the value somewhere else in
the builder struct? If not maybe we need to support skipping setter but
leaving a field to write. Can your use case be supported by validatiob in a
custom build method?

Dylan-DPC:
@lobisquit any updates on this?

lobisquit:
Hi, sorry but I currently have no time to work on the project that rised this issue, I think I will try to work it out in the future, but sadly not soon.

TedDriggs:
@lobisquit the best way to address this is probably to add something that says "I'm going to write my own setter". At that point, we could emit the field but not the setter.

mwilliammyers:
My use case for this is to add an item to a `Vec` before setting it. This cannot be accomplished as a `Default` because I want to add the item to the `Vec` regardless if the field was specified or not in the builder. This could be done with a custom builder, but would be much more ergonomic as a custom setter fn, because all the other fields can be handled normally.

Luro02:
In my use case I want to hide a field type from the user

```rust
#[derive(Builder)]
struct Example {
    #[builder(setter(skip))]
    data: BTreeMap<usize, OtherStruct>,
    // other fields omitted
}

impl ExampleBuilder {
    pub fn data(&mut self, values: Vec<OtherStruct>) -> &mut Self {
        let mut result = BTreeMap::new();

        for value in values {
            result.insert(value.number, value);
        }

        self.data = Some(result);
        self
    }
}
```

TedDriggs:
@Luro02 could you handle your use-case by doing the following?

1. Make the setter private
2. Change the setter name to `inner_your_field`
3. Expose the public setter manually

TedDriggs:
It looks like `setter(custom)` should handle this.

PR #154, opened 2019-11-02
setter(custom) which allows a custom setter implementation

jgraef:
Implements feature request #110

First I added an option to the `FieldLevelSetter` called `custom`. Such that the attribute is `#[builder(setter(custom))]`.

This option will behave like `skip` but not cause the field to be "skipped" (i.e. replaced with `PhantomData`). This is done by splitting the `enabled` flag into `setter_enabled` and `field_enabled`. `custom` will only cause the setter to be disabled, but not the field.

I'm aware that `custom` is missing for `StructLevelSetter`. To be honest, it's late and I need the feature. I can implement it tomorrow, if necessary :)

Tests are implemented in `tests/setter_custom.rs`.

I just started using this crate today and didn't look to deep into the code and it's capabilities (doc seems to be lacking). So there might be some inter-dependencies I overlooked. I copied the tests from `skip-setters.rs` and adapted them. I also added a test for `strip_option` as this might be a common usecase together with `custom`.

If there are any recommendations for improvements, let me know.

### DIFF
--- a/derive_builder/src/options/darling_opts.rs
+++ b/derive_builder/src/options/darling_opts.rs
@@ -116,13 +116,25 @@ pub struct FieldLevelSetter {
     into: Option<bool>,
     strip_option: Option<bool>,
     skip: Option<bool>,
+    custom: Option<bool>,
 }
 
 impl FieldLevelSetter {
-    /// Get whether or not this field-level setter indicates a setter should
-    /// be emitted. The setter shorthand rules are that the presence of a
-    /// `setter` with _any_ properties set forces the setter to be emitted.
-    pub fn enabled(&self) -> Option<bool> {
+    /// Get whether the setter should be emitted. The rules are the same as
+    /// for `field_enabled`, except we only skip the setter if `setter(custom)` is present.
+    pub fn setter_enabled(&self) -> Option<bool> {
+        if self.custom.is_some() {
+            return self.custom.map(|x| !x);
+        }
+
+        self.field_enabled()
+    }
+
+    /// Get whether or not this field-level setter indicates a setter and
+    /// field should be emitted. The setter shorthand rules are that the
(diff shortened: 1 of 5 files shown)


jgraef:
It just came to mind that the `strip_option` doesn't make sense in combination and the attribute can be omitted since the setter is custom anyway. I guess it doesn't hurt to have the test anyway.

jgraef:
I didn't follow up on the struct level attributes. I'm quite busy right now, but can do it at a later time.

I know this crate is kind of unmaintained or at least looking for a maintainer - which I don't get since builders are a thing that are widely used in the Rust ecosystem. How are the chances this get's merged soonish? We can just depend on a git for now, but would want to release our software to crates.io at some point, where we need a proper dependency.

I'm willing to offer some of my time to help maintain this project, if that's needed. Just a really quick feedback on this would be nice :)

jgraef:
Do I understand this correctly? It's a minor version bump because the interface of `FieldLevelSetter` changed? I didn't even consider that to be a public interface, but yeah of course it is.

But if I undo the renaming of the `enabled` method, this shouldn't require a minor version bump, right?

TedDriggs:
You're correct. However, I think your name is better so I've published with the bump. I'll file a separate issue to refactor in the future.

jgraef:
Nice.

A struct level attribute would also require a breaking change then. But does this attribute make sense on a struct level? The struct level attribute would basically disable all generated setter implementations. But you could always overwrite it with the field level attribute again.

Anyway, a refactor would be a good idea. I was also thinking about implementing a few other feature requests, since some of them seem pretty easy :) But I'll better wait for the refactoring then.

stoically:
Would be great to have this useful feature documented as I was only able to find it by searching here on github.

PR #169, opened 2020-07-24
Structural error

### LINKED ISSUE #60: structured errors instead of `String`
The `build` method currently returns `Result<Foo, String>`. But representing errors as `String` is not very rustic.

It's also a show-stopper for using `derive_builder` with `#![no_std]` on the stable toolchain. The reason is that `String` on `no_std` requires the `collections` crate, which is marked as unstable.

These are already two reasons that representing errors as `String` is not optimal.

# Proposal

Find a way to return structured errors instead of `Strings`. On `std` they should implement the usual `fmt` traits. On `no_std` they should be compliant with the stable toolchain.

# Questions

The main question is, how do we get the structured errors into scope? This is unfortunately _not_ straightforward, since `derive_builder` is a proc_macro crate.

> `proc-macro` crate types cannot export any items other than functions tagged with `#[proc_macro_derive]` currently.

So we need to work around that limitation somehow. These ideas come to my mind right now

1. We could add a `derive_builder_common` crate, which just exports traits / types / impls and must then be imported by any user of `derive_builder`.
2. We could add a flag to `derive_builder` to emit the traits / types / impls directly into the target crate. 
3. Each derivation with `derive_builder` could generate _unique_ error definitions for that struct.
4. Wait until proc_macros can export items.

All these approaches have their own drawbacks.
- 1+2 add boilerplate and complexity for users of `derive_builder`. The out-of-the-box experience is a bit impaired.
- 3 introduces unnecessary redundancy (and complexity). This might get ugly if a user has multiple derived builders and does pattern matching on the errors.
- 4 sounds like macros 2.0 which will most certainly take quite a while from now to land on stable.

cc @killercup

andy128k:
My attempt to resolve #60

### DIFF
--- a/derive_builder_core/src/builder.rs
+++ b/derive_builder_core/src/builder.rs
@@ -1,5 +1,5 @@
 use proc_macro2::TokenStream;
-use quote::{ToTokens, TokenStreamExt};
+use quote::{format_ident, ToTokens, TokenStreamExt};
 use syn::punctuated::Punctuated;
 use syn::{self, Path, TraitBound, TraitBoundModifier, TypeParamBound};
 
@@ -41,6 +41,38 @@ use Setter;
 /// pub struct FooBuilder {
 ///     foo: u32,
 /// }
+///
+/// #[doc="Error type for FooBuilder"]
+/// #[derive(Debug)]
+/// pub enum FooBuilderError {
+///     /// Uninitialized field
+///     UninitializedField(&'static str),
+///     /// Custom validation error
+///     ValidationError(String),
+/// }
+///
+/// impl ::std::convert::From<&'static str> for FooBuilderError {
+///     fn from(s: &'static str) -> Self {
+///         Self::UninitializedField(s)
+///     }
+/// }
+///
+/// impl ::std::convert::From<String> for FooBuilderError {
+///     fn from(s: String) -> Self {
+///         Self::ValidationError(s)
+///     }
+/// }
+///
+/// impl ::std::fmt::Display for FooBuilderError {
+///     fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+///         match self {
+///             Self::UninitializedField(ref field) => write!(f, "`{}` must be initialized", field),
+///             Self::ValidationError(ref error) => write!(f, "{}", error),
+///         }
+///     }
+/// }
+///
+/// impl ::std::error::Error for FooBuilderError {}
 /// #           ));
 /// #           #[cfg(not(feature = "clippy"))]
 /// #           result.append_all(quote!(#[allow(clippy::all)]));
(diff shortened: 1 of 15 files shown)


derive_builder_core/src/builder.rs
```diff
@@ -136,12 +168,47 @@ impl<'a> ToTokens for Builder<'a> {
...
+
             tokens.append_all(quote!(
                 #[derive(#derived_traits)]
                 #builder_doc_comment
                 #builder_vis struct #builder_ident #struct_generics #where_clause {
                     #(#builder_fields)*
                 }
+
+                #[doc=#builder_error_doc]
+                #[derive(Debug)]
```
TedDriggs, line 182: Does it make sense for this to also be marked `#[non_exhaustive]`? I know that would be an issue for older Rust versions, but not doing so would mean that changes to the enum in future would cause downstream crates to have to increment their versions.

derive_builder_core/src/builder.rs
```diff
@@ -136,12 +168,47 @@ impl<'a> ToTokens for Builder<'a> {
...
+                #builder_vis enum #builder_error_ident {
+                    /// Uninitialized field
+                    UninitializedField(&'static str),
+                    /// Custom validation error
+                    ValidationError(String),
+                }
+
+                impl ::std::convert::From<&'static str> for #builder_error_ident {
```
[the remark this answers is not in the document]
andy128k: Done

derive_builder_core/src/builder.rs
andy128k: Done

TedDriggs:
Thanks for submitting this. It's going to be a couple days before I can test it locally, but once I do I'll be able to provide the final review and will merge if everything looks good.

andy128k:
@TedDriggs any updates on this?

TedDriggs:
Sorry, my day job has been taking up a lot of my time and I haven't had a chance to test this. I think the code is all good, and the hangup is purely my "rustiness" with this code base. Here are my outstanding concerns:

1. I don't exactly remember how `derive-builder` handles the std-vs-nostd behavior, so I _think_ this is correct, but I'm not 100% sure
2. My ad-hoc inspection of builder errors suggests that they're primarily just having `unwrap` called on them, which leads me to think that this is safe. However, I don't know that I have a good way to test that theory, and even in `0.10.0` I'm loathe to break people's existing code
3. We're using `non_exhaustive` correctly, but I still feel bad about bumping rustc to 1.45.0 when moving to 0.10.0

Generating an error per builder is a much cleaner solution than trying to make one structured error that the main crate emits. That's definitely the right path forward.

I'm wondering if a good interim path forward is:

* Make using the error type an option passed to `#[builder(error(custom_enum))]` or something and ship that as 0.9.1. That would give early adopters the chance to use it now
* When we release `0.10.0`, make it the default but add a way to access the legacy string behavior

Thoughts?

andy128k:
@TedDriggs 

1. I do not have experience with `no_std`. It would be good to add automated tests for it.
2. I have an impression that people tend to `.unwrap()` or to `.map_err()`. This change doesn't break first use case. Those who `map_err()` either will welcome this change and tidy their code by removing `map_err` or will need to add `.to_string()`. While change is breaking, impact is not that big.
 I, personally decided to work on this issue because now I am one of the maintainers of `rss` crate whose users are asking for better error types.
3. Let me correct you. it is not 1.45.0 but 1.40.0 which was released 2019-12-19.

0.10.0-alpha/beta versions can be released for early adopters. SemVer helps here. Only those who use version `*` will be affected. So, potential harm and blast radius are small.

TedDriggs:
You are right. There used to be no_std tests, but I ended up disabling them because they broke too often. Let's move ahead with this on `0.10.0-alpha` and see how we go.

andy128k:
Conflicts are resolved.

TedDriggs:
I'm still thinking about the name of the variant for validation errors; I think I might switch it to `Custom` to follow what `serde` does. However, this is awesome work and I'm really grateful to you for bringing it in.

EndilWayfare:
My apologies, it's a little difficult for me to follow the macro shenanigans. It seems like the return type of the validator is now `Result<(), Into<#error_ty>>` instead of `Result<(), String>` (and `FooBuilder: From<String>` just an example implementation detail). Am I reading this correctly?

EndilWayfare:
Ok, wait, I'm sorry. I think I've got that wrong. `error_ty` _is_ the enum, and the `ValidationError` variant just has `String` data. What if there were a... `validation_error_ty` that defaults to `String`? I'd really prefer structured validation errors.

PR #174, opened 2020-08-12
extract derive macro into a separated crate

### LINKED ISSUE #155: Apply darling-style crate refactor
The current implementation of `derive_builder` exposes too much of the crate's internal behavior across the boundary between `derive_builder` and `derive_builder_core`. This results in unnecessary minor version bumps, as seen in #154.

andy128k:
Closes #155


andy128k:
@TedDriggs this is just a re-arrange of existing files for now.

derive_builder_macro/src/lib.rs
```diff
@@ -0,0 +1,80 @@
...
+    INIT_LOGGER.call_once(|| {
+        env_logger::init();
+    });
+
+    let ast = parse_macro_input!(input as syn::DeriveInput);
+
+    builder_for_struct(ast).into()
+}
+
+fn builder_for_struct(ast: syn::DeriveInput) -> proc_macro2::TokenStream {
```
TedDriggs, line 49: So this function should be moving to `derive_builder_core` and be made public; that way we can greatly restrict the surface area of the crate that exposes the main functionality. The macro crate then just calls that crate.

derive_builder_macro/CHANGELOG.md
```diff
@@ -0,0 +1,225 @@
+# Change Log
```
TedDriggs, line 1: Is this a symlink that GitHub isn't showing properly? I don't think we'll want to maintain a dedicated changelog for the macro crate.

andy128k: It is a copy. There is a dedicated changelog for core crate, so I did a copy for macro crate also.

andy128k:
@TedDriggs This refactoring allows to have single error type (placed in derive_builder crate) for all builders. Having such error type allows also to get rid of implicit conversions (&str -> UninitializedField, String -> ValidationError).

I am not sure what is the best direction. What do you think?

TedDriggs:
@andy128k we should continue to generate an error type for each builder. Unlike `serde` or `darling`, we don't want to publicly expose anything but the macro directly from this crate. The change here is mostly useful for allowing us to ensure we have durable references to `core` types.

andy128k:
@TedDriggs please review

TedDriggs:
@andy128k will do this weekend.

PR #175, opened 2020-08-19
Migrate from compiletest_rs to trybuild

andy128k:
I noticed that compiletest_rs is unstable. You need to run `cargo clean` from time to time to make it work. Its error messages are unreadable. Also I noticed that it doesn't work on macos.

Then I found [this](https://github.com/dtolnay/trybuild/issues/13).  :smirk:

I tested this on Linux and macos and it works well.

### DIFF
--- a/derive_builder/Cargo.toml
+++ b/derive_builder/Cargo.toml
@@ -32,8 +32,9 @@ derive_builder_core = { version = "=0.10.0-alpha", path = "../derive_builder_cor
 
 [dev-dependencies]
 pretty_assertions = "0.6"
-compiletest_rs = "0.5"
 skeptic = "0.13"
+rustversion = "1"
+trybuild = "1"
 
 [build-dependencies]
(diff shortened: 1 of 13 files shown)


andy128k:
[One more recommendation](https://github.com/dtolnay/trybuild/issues/84#issuecomment-670919391) how to test compiler failures.

andy128k:
@TedDriggs Rustc constantly changes/improves output of compile errors. Now tests in master branch are broken in nightly. Upcoming release of Rust will break tests on stable/beta too.

I propose to follow an [advise of dtolnay](https://github.com/dtolnay/trybuild/issues/84#issuecomment-670919391) and run them on a single version. Minimal supported version of compiler (for now it is 1.40.0) seems reasonable to me.

Switching to `trybuild` also improves developer's experience -- no need to do `cargo clean` between running tests.

andy128k:
@TedDriggs ping

PR #179, opened 2020-10-16
Fix requiring generic parameters to implement Default

ColonelThirtyTwo:
Currently the builder uses #[derive(Default)], which adds additional bounds to the type parameters
requiring them to implement Default. This is not required since the fields are all either `Option`
or `PhantomData`, which always have defaults, and means that Builders cannot be instantiated for
non-Default generics.

This commit makes the builder implement Default manually, without any additional bounds.

### DIFF
--- a/derive_builder_core/src/builder_field.rs
+++ b/derive_builder_core/src/builder_field.rs
@@ -75,6 +75,14 @@ impl<'a> ToTokens for BuilderField<'a> {
     }
 }
 
+impl<'a> BuilderField<'a> {
+    /// Emits a struct field initializer that initializes the field to `Default::default`.
+    pub fn default_initializer_tokens(&self) -> TokenStream {
+        let ident = self.field_ident;
+        quote!{ #ident : ::derive_builder::export::core::default::Default::default(), }
+    }
+}
+
 /// Helper macro for unit tests. This is _only_ public in order to be accessible
 /// from doc-tests too.
 #[doc(hidden)]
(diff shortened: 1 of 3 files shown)


derive_builder_core/src/builder_field.rs
```diff
@@ -75,6 +75,14 @@ impl<'a> ToTokens for BuilderField<'a> {
     }
 }
 
+impl<'a> BuilderField<'a> {
+    /// Emits a struct field initializer that initializes the field to `Default::default`.
+    pub fn default_initializer_tokens(&self) -> TokenStream {
+        let ident = self.field_ident;
+        quote!{ #ident : ::derive_builder::export::core::default::Default::default(), }
```
TedDriggs, line 82: Is there a reason to use `Default::default()` here instead of `::derive_builder::export::core::option::Option::None`?

TedDriggs (commented):
I'm not going to try and unwind the Travis problems before merging this, so once the question in the review is answered it should be ready to go into master.

derive_builder_core/src/builder_field.rs
ColonelThirtyTwo: I don't remember. Maybe it was to work with optional fields that aren't wrapped in `None`.

TedDriggs: Ah, I think it's for the `PhantomData` case you mentioned.

PR #188, opened 2021-01-13
Remove skeptic

TedDriggs:
Complexity in the repo's tests have made it difficult for maintainers
and contributors to work on the crate with confidence that they haven't
caused regressions. `skeptic` was a major driver of that complexity, as
it introduced the need for an additional piece of the test matrix and
custom build scripts.

That complexity was only testing a single example in the README, and
that example is already covered by the examples/ directory. Removing
skeptic simplifies the repo and the CI test pass with minimal reduction
in test coverage.


TedDriggs:
@andy128k can you take a look at this and see if I've missed anything as far as 1) where `skeptic` was adding value or 2) where `skeptic` had a footprint in the repo?

andy128k:
@TedDriggs Probably whole module `derive_builder_core::doc_tpl` should be removed too. Looks like it is not used anymore.

Issue #191, opened 2021-01-13
Allow specification of custom error type


TedDriggs:
Crates may already have an error into which they'd like to feed builder failures. Add `#[builder(build_fn(error = "..."))]` to allow specifying that, and document the contract for that custom error type.

andy128k:
Alternatively a simpler solution is possible. Just allow to make `build()` method and error type private.

```rust
#[derive(Builder)]
#[builder(error(private))]
#[builder(build_fn(name = "build_impl", private))]
struct Lorem {
    pub ipsum: u8,
}

impl LoremBuilder {
    pub fn build(&self) -> Result<Lorem, CustomError> {
        self.build_impl()
            .map_err(|private_lorem_builder_error| CustomError::new(... private_lorem_builder_error...))
    }
}
```

Issue #56, opened 2017-03-23
Infallible Build method


colin-kiegel:
The build method currently fails, if a field value was not initialized and no default value was provided (possible since custom defaults #3 are implemented).

# Proposal

Add a way to change the signature of the `build` method from `Result<Foo, _>` to `Foo`. This method would _not_ panic. Instead `derive_builder` would deny that change of signature if the method is fallible.

# Questions

How should this annotation be called? Here are two suggestions:
- `#[builder(infallible)]`
- `#[builder(no_result)]`
- `#[builder(deny_failure)]`

Should this be scoped with other build-method-related options, e.g. inside - `#[builder(buildfn(...))]`?

# Outlook

It is likely that `derive_builder` will have other failure modes in the future, such as field-specific custom validations (`foo` must be in `0..10`, etc.). These features can not be activated if the infallible build method is also active.

# Alternatives

- Change the signature of the `build` method automatically when it cannot fail. This may be surprising if you have a lot of code depending on the current signature, then you add a default value to a field (which seems like a non-breaking change) and then nothing compiles anymore. Not good.
- Always allow changing the signature to `Foo` even if the `build` method is fallible. Panic on failure. This does not seem to be idiomatic Rust - which we try to achieve as best as possible. If someone really wants that behaviour he/she can implement a custom wrapper (we will add means to rename the build method and make it private).

colin-kiegel:
Anyone reading this - I am looking for feedback on the naming question. This should be reasonably understandable to anyone reading it for the first time, without prior knowledge of `derive_builder`.

A general question is, whether this option should be scoped with other settings that refer to the `build()` method, like skipping, renaming, visibility-level (public/private). I will add this question, but it seems like scoping would be a good idea.

TedDriggs:
> Always allow changing the signature to Foo even if the build method is fallible. Panic on failure. This does not seem to be idiomatic Rust - which we try to achieve as best as possible. If someone really wants that behaviour he/she can implement a custom wrapper (we will add means to rename the build method and make it private).

I'd argue that this _is_ idiomatic; Rust often provides methods which panic on failure, and leaves it up to the programmer to decide whether to use those methods or to use the ones that return `Option` or `Result`. A great example is the `Index` trait vs. the `get` method; the former panics while the latter returns an option.

[...]

PR #194, opened 2021-01-15
Custom error example

TedDriggs:
@andy128k this is an interesting example of using custom validation errors today. For #181, we'd need a way for this to work without the heap allocation of `String`, which is what's driving #191.

### DIFF
--- a/derive_builder_core/src/error.rs
+++ b/derive_builder_core/src/error.rs
@@ -1,0 +1,32 @@
+use std::{error::Error, fmt};
+
+/// Runtime error when a `build()` method is called and one or more required fields
+/// do not have a value.
+#[derive(Debug, Clone)]
+pub struct UninitializedFieldError(&'static str);
+
+impl UninitializedFieldError {
+    /// Create a new `UnitializedFieldError` for the specified field name.
+    pub fn new(field_name: &'static str) -> Self {
+        UninitializedFieldError(field_name)
+    }
+
+    /// Get the name of the first-declared field that wasn't initialized
+    pub fn field_name(&self) -> &'static str {
+        self.0
+    }
+}
+
+impl fmt::Display for UninitializedFieldError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Field not initialized: {}", self.0)
+    }
+}
+
+impl Error for UninitializedFieldError {}
+
+impl From<&'static str> for UninitializedFieldError {
+    fn from(field_name: &'static str) -> Self {
+        Self::new(field_name)
+    }
+}
--- a/derive_builder/examples/custom_error.rs
+++ b/derive_builder/examples/custom_error.rs
@@ -12,7 +12,7 @@ use std::fmt;
 fn validate_age(builder: &ExampleBuilder) -> Result<(), Error> {
     match builder.age {
         Some(age) if age > 150 => Err(Error::UnrealisticAge(age)),
-        _ => Ok(())
+        _ => Ok(()),
     }
 }
 
--- a/derive_builder/examples/custom_error.rs
+++ b/derive_builder/examples/custom_error.rs
@@ -1,0 +1,52 @@
+//! This example shows using custom validation with a non-string error type.
+//!
+//! This relies on how the generated build function is constructed; the validator
+//! is invoked in conjunction with the `?` operator, so anything that converts to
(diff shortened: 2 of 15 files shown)


andy128k:
@TedDriggs I am afraid of the situation when user will provide different non-compatible parameters and will receive a cryptic error message.

```rust
#[derive(Builder)]
#[builder(error = CError, unknown_field = "CError::WTF")]
struct TheStruct {
   #[builder("self.a_default()?")]
   a: i32,
   #[builder("self.b_default()?")]
   b: i32,
}

impl TheStructBuilder {
    fn a_default() -> Result<i32, AError> { ... }
    fn b_default() -> Result<i32, BError> { ... }
}

enum CError {
    WTF(&'static str),
    A(AError),
    B(BError),
}

impl From<AError> for CError { ... } // missing this will cause a cryptic error "From<AError> is not implemented in CError. Error is inside a macro"
impl From<BError> for CError { ... }
```

This looks like an attempt to create a new language inside of attributes.
`"self.a_default()?"` are already "unstructured" AF (sorry for my language).

TedDriggs:
For the record, I feel passing a block in `default` is an anti-pattern, and were it not for backwards compatibility I would have gone with the serde pattern of accepting a `path` in that position. We can improve the confusing errors situation there by making the emitted spans better, but in general it's not something I'd encourage people to use.

andy128k:
@TedDriggs This crate has version 0.9. Semver allows to break compatibility. Let's change default attribute. Users who use `?` may be advised to move validation into a custom validator.

andy128k:
Side note: Requiring `From<derive_builder::UninitializedFieldError>` for custom error may contradict (or clash) with "infallible" builders #56 (which I would like this crate to implement in a future 😏 ).

derive_builder_core/src/error.rs
```diff
@@ -0,0 +1,32 @@
...
+
+impl fmt::Display for UninitializedFieldError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "Field not initialized: {}", self.0)
+    }
+}
+
+impl Error for UninitializedFieldError {}
+
+impl From<&'static str> for UninitializedFieldError {
```
andy128k, line 28: Is this really needed?

TedDriggs:
@andy128k the macro itself doesn't - indeed, it can't - require that the specified error type have the impl; it can only attempt to use the impl and fail if it's not present. Builders with default values technically don't require this `impl` as it stands today; I've added a run-pass test to make sure that stays the case.

derive_builder_core/src/error.rs
TedDriggs: I suppose not; is there a reason to exclude it? I generally include `From` conversions when `new` takes a single argument, but the macro could invoke `new` directly.

andy128k: Same concerns as in #192. `&str` is too broad.

TedDriggs: In this case there isn't ambiguity though; we know that we intended to represent an uninitialized field, as opposed to a custom validation error. The macro-generated code uses `UninitializedFieldError::from` since `.into().into()` almost never works in Rust.

andy128k:
Possible option is to add additional parameter.

```
#[builder(default)] // uses Default::default()
#[builder(default = "42")] // uses "42"
#[builder(default_method = method_ident)] // uses self.method_ident
```

Downside: more changes for those who migrate from 0.9.

TedDriggs:
Feel free to create an issue for the `default` question; I don't think it blocks this PR, since the advanced scenario for that was pretty bad before this PR too.

At this point, I believe the following are true:

1. This PR allows the `validate` function to return `&'static str` without it being mistaken for an uninitialized field.
2. This PR allows the creation of errors that don't allocate, even with `validate` being used.
3. When a conversion is missing for uninitialized field or for `validate`, the compile error produced is readable.
4. This PR does not require any new crates to directly depend on `derive_builder`.

As long as these hold, I'm inclined to move forward with this approach instead of the approach in #192.

andy128k:
As soon as this PR removes `From<&str> for #builder_error_ident`, I am happy to close my PR in favor of this one.

derive_builder_core/src/error.rs
TedDriggs: I've confirmed that there is no `From<&'static str>` direct to the generated builder error, only to the `UninitializedFieldError`.

Issue #114, opened 2018-03-31
Special setter behavior for `Vec` fields?


pwoolcoc:
Would it be possible to have special setter methods for `Vec` fields? That is, instead of just a field for setting the whole vec at once, the user could opt-in to also generating setters that `.push` an element of type `T` onto a field with type `Vec<T>`?

The generated code could look something like

```
struct FooBuilder {
    bars: Option<Vec<Bar>>
}

impl FooBuilder {
    ...
    pub fn bars<VALUE: Into<Vec<Bar>>>(&mut self, value: VALUE) -> &mut Self {
        self.bars = Some(value.into());
        self
    }
    pub fn bar<VALUE: Into<Bar>>(&mut self, value: VALUE) -> &mut Self {
        if let Some(ref mut bars) = self.bars {
            bars.push(value.into());
        } else {
            self.bars = Some(vec![value.into()]);
        }
        self
    }
    ...
}
```

Ideally I'd like to have this for any `.push`-able type, but until we have collections traits I think having it just for `Vec` could be helpful.

TedDriggs:
Conceptually it's doable. I don't think it'd be possible for the names to be automatically generated to be singular or plural, but other than that it doesn't seem too terrible. What are the cases you have in mind for this feature?

TedDriggs:
FWIW, I have some experience implementing this in [`darling`](https://github.com/TedDriggs/darling/blob/master/core/src/codegen/field.rs#L86). It wasn't too bad, but I was disappointed to find that there isn't a good trait to latch onto which would allow arbitrary types to be used.

carloscasalar:
It could by like [@Singular annotation in lombok](https://projectlombok.org/features/Builder) specifying   the singular method name. I Think it could be something like this:

```rust
#[derive(Default, Builder, Debug)]
struct FooBuilder {
    #[builder(singular(name = "bar"))]
    bars: Option<Vec<Bar>>
}
```

cheako:
Don't leave out hashmap, probably implement something for all the std collections.

cheako:
I'm writing a game engine like TheCherno's Hazel.  He has this habit of creating a class, calling a setter on it and then passing the results to another class's setter.

https://github.com/TheCherno/Hazel/blob/30b5c242986c67ab20709980f88ca115e65e1261/Sandbox/src/SandboxApp.cpp#L25

30,31,36

[...]

PR #196, opened 2021-01-22
Add an option to extend collection-like fields

andy128k:
Addresses #114

### DIFF
--- a/derive_builder/tests/setter_extend.rs
+++ b/derive_builder/tests/setter_extend.rs
@@ -1,0 +1,43 @@
+#[macro_use]
+extern crate pretty_assertions;
+#[macro_use]
+extern crate derive_builder;
+
+use std::collections::HashMap;
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+struct Lorem {
+    #[builder(extend)]
+    foo: String,
+    #[builder(extend)]
+    bar: Vec<String>,
+    #[builder(extend)]
+    baz: HashMap<String, i32>,
+}
+
+#[test]
+fn generic_field() {
+    let x = LoremBuilder::default()
+        .foo("foo".into())
+        .bar_extend_one("bar".into())
+        .bar_extend_one("bar bar".into())
+        .bar_extend_one("bar bar bar".into())
+        .foo_extend_one('-')
+        .baz_extend_one(("baz".into(), 1))
+        .baz_extend_one(("bazz".into(), 2))
+        .baz_extend_one(("bazzz".into(), 3))
+        .foo_extend_one("foo")
+        .build()
+        .unwrap();
+
+    assert_eq!(
+        x,
+        Lorem {
+            foo: "foo-foo".into(),
+            bar: vec!["bar".into(), "bar bar".into(), "bar bar bar".into()],
+            baz: vec![("baz".into(), 1), ("bazz".into(), 2), ("bazzz".into(), 3)]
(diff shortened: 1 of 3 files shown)


TedDriggs (commented):
Thanks for putting this together, and great work on the code. Do we have examples of analogous builder APIs from crates that hand-implement their builders? The syntax for this feels a bit strange, but that might just be that it's new to me.

andy128k:
@TedDriggs I am not aware of such crates. I agree that name of setter `<field>_extend_one` is bumpy (it follows [std](https://doc.rust-lang.org/stable/std/iter/trait.Extend.html#method.extend_one)). But guessing singular name of a setter from the plural name of a field and do it properly (English is hard :)) will take much more code than this feature deserves.

- `<field>_append` or `append_<field>` look better but semantically incorrect for unordered collections (e.g. `HashMap)`
- `<field>_insert` or `insert_<field>` are too broad (what is an insert position?)
- `<field>_add`  or `add_<field>`  are too Java-ish :smirk:
- Here my English hardcover reaches a hardcover.

TedDriggs:
> Here my English vocabulary reaches a hardcover.

Your vocabulary is totally up to the task; the challenge here is that the underlying concept is more complex than a single word will accurately convey.

In the original issue, one proposal is that for some field `example`, calling `builder.example(...)` would push/extend/add an item, while `builder.examples(...)` would overwrite the whole collection. What if the declaration looked like this:

```rust
pub struct MyStruct {
    #[builder(setter(multiple(replace = "policies")))]
    policy: Vec<Policy>,
}
```

This would then generate

```rust
pub struct MyStructBuilder {
    policy: Option<Vec<Policy>>,
}

pub struct MyStructBuilder {
    policy: Option<Vec<Policy>>,
}

impl MyStructBuilder {
    pub fn policy(&mut self, value: Policy) -> &mut Self {
        self.policy
            .get_or_insert_with(|| Default::default())
            .extend(::std::iter::once(policy));
        self
    }

    pub fn policies(&mut self, value: Vec<Policy>) -> &mut Self {
        self.policy = Some(value);
        self
    }
}
```

The second setter is only emitted if `replace = "..."` is present in the setter's declaration. The main advantage of this approach is that the names stay concise, and each additional setter is opt-in:

* The simplest `builder` attribute provides the simplest behavior
* Declaring `#[builder(setter(multiple))]` changes the setter using the `extend` approach you've shown here, but doesn't force a second setter
* Declaring `#[builder(setter(multiple(replace = "...")))]` does create both setters, and puts the burden on the deriver to choose a good name.

Thoughts?

andy128k:
@TedDriggs  What if a field is named in plural form? In the case of `#[builder(setter(multiple))]` a setter will be `fn policies(&mut self, value: Policy)`? This looks strange.

What about `#[builder(setter(multiple(one = "policy", all = "policies")))]`? And if names are not given (`#[builder(setter(multiple))]` then fall back to `<field_name>` for all and `<field_name>_add` for one.

ShaneEverittM:
In David Tolnay's [proc-macro workshop](https://github.com/dtolnay/proc-macro-workshop#derive-macro-derivebuilder), he uses the syntax: `#[builder(each = "name")]` above the field to be extended, which will generate a function called "name" that extends the collection.

andy128k:
I like `each = "name"` and created an alternative PR #199

TedDriggs:
Closing in favor of #199

PR #199, opened 2021-03-15
Add an option to extend collection-like fields (`each`-version)

### LINKED ISSUE #114: Special setter behavior for `Vec` fields?

andy128k:
Closes #114 
Closes #196

### DIFF
--- a/derive_builder/tests/setter_extend.rs
+++ b/derive_builder/tests/setter_extend.rs
@@ -1,0 +1,94 @@
+#[macro_use]
+extern crate pretty_assertions;
+#[macro_use]
+extern crate derive_builder;
+
+use std::collections::HashMap;
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+struct Lorem {
+    #[builder(setter(each = "foo_append"))]
+    foo: String,
+    #[builder(setter(each = "bar"))]
+    bars: Vec<String>,
+    #[builder(setter(each = "baz"))]
+    bazes: HashMap<String, i32>,
+}
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+#[builder(pattern = "mutable")]
+struct Ipsum {
+    #[builder(setter(each = "foo_append"))]
+    foo: String,
+    #[builder(setter(each = "bar"))]
+    bars: Vec<String>,
+    #[builder(setter(each = "baz"))]
+    bazes: HashMap<String, i32>,
+}
+
+#[test]
+fn extend_field() {
+    let x = LoremBuilder::default()
+        .foo("foo".into())
+        .bar("bar".into())
+        .bar("bar bar".into())
+        .bar("bar bar bar".into())
+        .foo_append('-')
+        .baz(("baz".into(), 1))
+        .baz(("bazz".into(), 2))
+        .baz(("bazzz".into(), 3))
+        .foo_append("foo")
+        .build()
+        .unwrap();
+
+    assert_eq!(
+        x,
+        Lorem {
+            foo: "foo-foo".into(),
+            bars: vec!["bar".into(), "bar bar".into(), "bar bar bar".into()],
+            bazes: vec![("baz".into(), 1), ("bazz".into(), 2), ("bazzz".into(), 3)]
+                .into_iter()
+                .collect(),
+        }
+    );
+}
+
+#[test]
+fn extend_field_mutable() {
+    let x = IpsumBuilder::default()
+        .foo("foo".into())
+        .bar("bar".into())
+        .bar("bar bar".into())
+        .bar("bar bar bar".into())
+        .foo_append('-')
+        .baz(("baz".into(), 1))
+        .baz(("bazz".into(), 2))
+        .baz(("bazzz".into(), 3))
+        .foo_append("foo")
+        .build()
+        .unwrap();
+
+    assert_eq!(
+        x,
+        Ipsum {
+            foo: "foo-foo".into(),
+            bars: vec!["bar".into(), "bar bar".into(), "bar bar bar".into()],
+            bazes: vec![("baz".into(), 1), ("bazz".into(), 2), ("bazzz".into(), 3)]
+                .into_iter()
+                .collect(),
+        }
+    );
+}
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+#[builder(setter(skip))]
+struct Dolor {
+    #[builder(setter(each = "foo"))]
+    foos: Vec<i32>,
+}
+
+#[test]
+fn extend_field_enabled() {
+    let x = DolorBuilder::default().foo(1).foo(2).build().unwrap();
+    assert_eq!(x, Dolor { foos: vec![1, 2] });
+}
--- a/derive_builder_core/src/macro_options/darling_opts.rs
+++ b/derive_builder_core/src/macro_options/darling_opts.rs
@@ -130,6 +130,7 @@ pub struct FieldLevelSetter {
     strip_option: Option<bool>,
     skip: Option<bool>,
     custom: Option<bool>,
+    each: Option<Ident>,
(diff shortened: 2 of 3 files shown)


andy128k:
This implementation does not support things like `#[builder(setter(skip = "true", each = "foo"))]`.

TedDriggs:
This feels like the right API; thanks for taking the time to go through multiple iterations on it.

> This implementation does not support things like `#[builder(setter(skip = "true", each = "foo"))]`.

What happens if someone tries? Does the `skip` preempt the `each`, or does the `each` overrule the `skip`?

derive_builder_core/src/setter.rs
```diff
@@ -151,7 +154,26 @@ impl<'a> ToTokens for Setter<'a> {
...
+                    #(#attrs)*
+                    #[allow(unused_mut)]
+                    #vis fn #ident_each <ITEM>(#self_param, item: ITEM) -> #return_ty
+                    where
+                        #ty: ::derive_builder::export::core::default::Default + ::derive_builder::export::core::iter::Extend<ITEM>,
+                    {
+                        #deprecation_notes
+                        let mut new = #self_into_return_ty;
+                        new.#field_ident
+                            .get_or_insert_with(|| ::derive_builder::export::core::default::Default::default())
```
TedDriggs, line 172:
```suggestion
                            .get_or_insert_with(::derive_builder::export::core::default::Default::default)
```
This would satisfy the [redundant_closure](https://rust-lang.github.io/rust-clippy/v0.0.212/#redundant_closure) clippy lint, but I've seen cases in the past where that lint would recommend invalid code so you should probably double-check the change doesn't break anything.

derive_builder_core/src/setter.rs
```diff
@@ -151,7 +154,26 @@ impl<'a> ToTokens for Setter<'a> {
...
-                }));
+                    }
+                ));
+            }
+
+            if let Some(ref ident_each) = self.each {
+                tokens.append_all(quote!(
+                    #(#attrs)*
+                    #[allow(unused_mut)]
+                    #vis fn #ident_each <ITEM>(#self_param, item: ITEM) -> #return_ty
```
TedDriggs, line 165: Should this honor `setter(into)`, or is that going to cause type inference problems trying to figure out `ITEM`?

TedDriggs, line 165: Please use `VALUE` instead of `ITEM` here so that we continue to only reserve one type parameter name for the macro's own use.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -130,6 +130,7 @@ pub struct FieldLevelSetter {
     strip_option: Option<bool>,
     skip: Option<bool>,
     custom: Option<bool>,
+    each: Option<Ident>,
```
TedDriggs, line 133: Related to the comment about `skip`, it looks like `each` is _not_ being considered in [this check](https://github.com/colin-kiegel/rust-derive-builder/pull/199/files#diff-fce05abe2a1a48a344afbebd38b64914fde1d4131faa01fa9d32e5804b6f31eeR151) for `field_enabled`. That seems like an oversight.

derive_builder_core/src/setter.rs
```diff
@@ -151,7 +154,26 @@ impl<'a> ToTokens for Setter<'a> {
...
+                    where
+                        #ty: ::derive_builder::export::core::default::Default + ::derive_builder::export::core::iter::Extend<ITEM>,
+                    {
+                        #deprecation_notes
+                        let mut new = #self_into_return_ty;
+                        new.#field_ident
+                            .get_or_insert_with(|| ::derive_builder::export::core::default::Default::default())
+                            .extend(::derive_builder::export::core::option::Option::Some(item));
```
TedDriggs, line 173: Is `Some` being used here instead of `iter::once`?

TedDriggs (changes requested):
Some minor changes needed, but this is on track to be merged and go out as part of 0.10.0

andy128k:
> This feels like the right API; thanks for taking the time to go through multiple iterations on it.
> 
> > This implementation does not support things like `#[builder(setter(skip = "true", each = "foo"))]`.
> 
> What happens if someone tries? Does the `skip` preempt the `each`, or does the `each` overrule the `skip`?

No setter is generated and `each` is ignored.

derive_builder_core/src/setter.rs
andy128k: Yes.

BTW, `iter::once` is implemented like this:

```
pub fn once<T>(value: T) -> Once<T> {
    Once { inner: Some(value).into_iter() }
}
```

:)

derive_builder_core/src/setter.rs
andy128k: Done

derive_builder_core/src/setter.rs
andy128k: Done

derive_builder_core/src/setter.rs
andy128k: Ideally, I would use `extend_one` here, but it is not stabilized yet. So, I just [inlined it](https://doc.rust-lang.org/stable/src/core/iter/traits/collect.rs.html#339-342).

derive_builder_core/src/setter.rs
andy128k: Yeah, type inference doesn't work well here.

Additionally it requires to introduce one more type parameter `VALUE: Into<ACTUAL_VALUE>`. Type parameter may be replaced by an `impl Trait` approach, but, IIRC, this crate still targets 2015 edition, so this will not work either.

derive_builder_core/src/macro_options/darling_opts.rs
andy128k: Fixed. And corresponding test is added.

derive_builder/tests/setter_extend.rs
```diff
@@ -0,0 +1,94 @@
...
+                .into_iter()
+                .collect(),
+        }
+    );
+}
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+#[builder(setter(skip))]
+struct Dolor {
+    #[builder(setter(each = "foo"))]
```
TedDriggs, line 86: Final question: If I write

```rust
struct Dolor {
    #[builder(setter(skip, each = "foo"))]
    foos: Vec<i32>,
}
```

will I get the each form without also getting the set-all form?

TedDriggs (approved):
See final question; otherwise this looks ready to merge

derive_builder/tests/setter_extend.rs
andy128k: It behaves consistently with other attributes (`skip` overrides all of them). So, answer is `no`, no methods will be generated.

That's because generation of `each` methods is nested into a generator of a field setter.

```
if field_enabled {
    // generate setter
    if let Some(each) = each {
        // generate each setter
    }
}
```

PR #204, opened 2021-04-13
Make it work in no_std environment

### LINKED ISSUE #201: Can't compile v0.10.0 in the `no_std` environment

andy128k:
Closes #201

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,3 +49,9 @@ jobs:
           cargo build ${{ matrix.features }} --verbose
           cargo test --no-fail-fast ${{ matrix.features }} --verbose -- --nocapture
           cargo doc
+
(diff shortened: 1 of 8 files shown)


TedDriggs:
What happens in environments where allocations are impossible? Would this compile and then fail if someone tried to construct that variant?

andy128k:
It will not compile. The same as it is in `0.9.0`.

PR #206, opened 2021-04-21
Move UninitializedFieldError to derive_builder

### LINKED ISSUE #201: Can't compile v0.10.0 in the `no_std` environment

TedDriggs:
UninitializedFieldError is part of the runtime API of derive_builder,
not part of its compile-time macro machinery. Exporting it from the
derive_builder_core crate caused downstream crates to unnecessarily
import all of derive_builder_core, rather than only importing it via
the proc-macro crate at compile-time.

Fixes #201


TedDriggs:
@andy128k any idea why this fails when the original PR adding the test succeeds? I'm very unfamiliar with `no_std`.

TedDriggs:
Filed an issue to bring back the removed test, so proceeding to merge.

PR #208, opened 2021-04-22
Re-introduce test crate for no_std environment

### LINKED ISSUE #207: Reintroduce compile-time tests for no_std
In #206 we removed the `no_std` test added in #204 to be able to publish 0.10.2. We should reintroduce the test to make sure we don't break `no_std` unexpectedly going forward.

Note that this is the second time we've had to remove a `no_std` test suite because it was failing too often for reasons that are outside the scope of this crate.

andy128k:
Closes #207

### DIFF
--- a/derive_builder_no_std_tests/Cargo.toml
+++ b/derive_builder_no_std_tests/Cargo.toml
@@ -1,0 +1,11 @@
+[package]
+name = "derive_builder_no_std_tests"
+version = "0.1.0"
+authors = ["Andrey Kutejko <andy128k@gmail.com>"]
+edition = "2018"
+
+[features]
+clippy = ["derive_builder/clippy"]
+
+[dependencies]
+derive_builder = { path = "../derive_builder", default-features = false }
--- a/derive_builder_no_std_tests/Cargo.toml
+++ b/derive_builder_no_std_tests/Cargo.toml
@@ -3,6 +3,7 @@ name = "derive_builder_no_std_tests"
 version = "0.1.0"
 authors = ["Andrey Kutejko <andy128k@gmail.com>"]
 edition = "2018"
+publish = false
 
 [features]
 clippy = ["derive_builder/clippy"]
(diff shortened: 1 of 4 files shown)


derive_builder_no_std_tests/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "derive_builder_no_std_tests"
+version = "0.1.0"
+authors = ["Andrey Kutejko <andy128k@gmail.com>"]
+edition = "2018"
```
TedDriggs, line 5:
```suggestion
edition = "2018"
publish = false
```

This will prevent the package from being published accidentally ([source](https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field)).

TedDriggs (approved):
I don't see what's different in this PR from yesterday, but it's passing CI so maybe the failure was just a fluke?

TedDriggs:
Will merge after CI passes.

andy128k:
@TedDriggs  The difference is huge! It was a binary crate and now it is a library crate.

What actually happened is that the crate was linked to the `libstd`. When you fixes the issue with a dependencies you also removed linkage of the crate with `libstd` and it broke. (This is an additional confirmation that your fix actually removed a dependency on `libstd`).

Now this crate is a lib (not a binary) and there is no linking process at all (excluding building of tests, but they imply `libstd` anyway).

Also, this "test" is less strict than the previous one. We may resurrect the old one (binary), but it will work on nightly only :(

TedDriggs:
If you go back in the repo history you'll see that there was such a test at one point. Then it broke due to changes in nightly, nobody knew how to fix it, and eventually I deleted it because it hadn't worked in ages and was adding a bunch of complexity to the tests beyond `cargo test --all`.

Let's see if we have further problems with `no_std`; now that we know what to look out for, I think we'll be safe.

PR #214, opened 2021-07-19
Allow collection setters to be generic over `Into`

### LINKED ISSUE #209: Allowing `each` and `into` on a vec
Hello,

I'm trying to have a nice api for one of my crates and I was wondering if I could be able to have a struct like this:

```rust

#[derive(derive_builder::Builder, Debug, Clone)]
#[builder(setter(into))]
struct SessionManagementPacketResult {
    #[builder(default = "SessionState::Opening")]
    session_state: SessionState,
    #[builder(setter(each = "packet", into))]
    packets: Vec<Packet>,
}
```

I want to be able to do `session.packet(xx).packet(xx)` where `xx` implement `From<xx> for Packet` (the into part) but it seems to not work, or the `into` isn't applied to `packet`.

Any idea of a workaround on that? 

Thank's!

uint:
Apparently this works without doing any special trickery, at least on my version of the compiler.

Closes #209

### DIFF
--- a/derive_builder_core/src/setter.rs
+++ b/derive_builder_core/src/setter.rs
@@ -162,10 +162,24 @@ impl<'a> ToTokens for Setter<'a> {
             if let Some(ref each) = self.each {
                 let ident_each = &each.name;
 
+                let ty_params: TokenStream;
+                let param_ty: TokenStream;
+                let into_item: TokenStream;
+
+                if each.into {
+                    ty_params = quote!(<VALUE, FROM_VALUE: ::derive_builder::export::core::convert::Into<VALUE>>);
+                    param_ty = quote!(FROM_VALUE);
+                    into_item = quote!(item.into());
+                } else {
+                    ty_params = quote!(<VALUE>);
+                    param_ty = quote!(VALUE);
+                    into_item = quote!(item);
+                }
+
                 tokens.append_all(quote!(
                     #(#attrs)*
                     #[allow(unused_mut)]
-                    #vis fn #ident_each <VALUE>(#self_param, item: VALUE) -> #return_ty
+                    #vis fn #ident_each #ty_params(#self_param, item: #param_ty) -> #return_ty
                     where
                         #ty: ::derive_builder::export::core::default::Default + ::derive_builder::export::core::iter::Extend<VALUE>,
                     {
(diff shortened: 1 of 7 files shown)


TedDriggs:
Thanks for submitting this; I haven't had the time to review it yet. @andy128k I believe you were the original author; can you take a look at these changes please?

derive_builder_core/src/setter.rs
```diff
@@ -158,19 +159,35 @@ impl<'a> ToTokens for Setter<'a> {
...
+                let ident_each = &each.name;
+
+                let ty_params: TokenStream;
+                let param_ty: TokenStream;
+                let into_item: TokenStream;
+
+                if each.into {
+                    ty_params = quote!(<VALUE, FROM_VALUE: ::derive_builder::export::core::convert::Into<VALUE>>);
+                    param_ty = quote!(FROM_VALUE);
+                    into_item = quote!(item.into());
```
andy128k, line 172: `into_item = quote!(::derive_builder::export::core::convert::Into::into(item));`

andy128k (commented):
This will be a breaking change, so a minor version bump is needed.

I would also like to see how/if this works with maps (`HashMap`/`BTreeMan`). I suppose maps with a `String` key are pretty common and being able to use `str` literals would be a huge improvement.

uint:
> I would also like to see how/if this works with maps (`HashMap`/`BTreeMan`). I suppose maps with a `String` key are pretty common and being able to use `str` literals would be a huge improvement.

I think for this to Just Work there would have to be some generic thing of this sort in the std:

```rs
impl<T, U, FromT: Into<T>, FromU: Into<U>> From<(FromT, FromU)> for (T, U) { ... }
```

Or maybe something like this?
```rs
impl<T, U, FromT: Into<T>, FromU: Into<U>> Extend<(FromT, FromU)> for HashMap<T, U> { ... }
```

But there is not. The best I can think of is adding a separate `into_pair` syntax, like so:

```rs
#[builder(setter(each(name = "bar", into_pair)))]
```

On the bright side, this would let us avoid having to pass a tuple. We could make the function accept two arguments - key and value.

andy128k (approved):

TedDriggs:
I'm going to rebase this and address the conflicts. Given that `each` has been in the wild for some time now, I'm inclined to create a wrapper for `Each` that will enable it to read the current `each = "..."` form so there's no breaking change; `#[builder(setter(each = "..."))]` would be the same as `#[builder(setter(each(name = "....")))]. @uint / @andy128k thoughts on that?

andy128k:
Frankly, I don't like how it becomes lisp-ish with all those nested constructs, so I welcome idea to continue to support current (relatively) flat syntax.

TedDriggs:
I've updated the branch [here](https://github.com/colin-kiegel/rust-derive-builder/tree/each-into), but had some difficulty with `strip_option` and `Option` wrappers around collections.

TedDriggs:
This merged as #234

PR #229, opened 2022-02-24
Preserve spans for default = "..." metas

### LINKED ISSUE #195: Improve spans for default blocks
Right now, bad code in `default = "..."` expressions don't produce useful error spans. It would be an improvement to point to the offending block.

TedDriggs:
When a user provides an explicit default = "..." expression,
errors in that expression pointed to the macro call-site.
This made it difficult to determine which expression was responsible.

This commit changes how the block is built, preserving the span of the input literal string.

Fixes #195


derive_builder_core/src/block.rs
```diff
@@ -27,25 +27,32 @@ use syn;
...
-            { #( #inner )* }
-        ));
+        self.0.to_tokens(tokens)
     }
 }
 
-impl FromStr for Block {
-    type Err = String;
+#[cfg(test)]
+impl std::str::FromStr for BlockContents {
```
TedDriggs, line 54: This should be removed in favor of a helper function in the tests.

derive_builder_core/src/block.rs
```diff
@@ -55,26 +62,47 @@ impl FromStr for Block {
...
+    fn try_from(s: &LitStr) -> Result<Self, Self::Error> {
+        let mut block_str = s.value();
+        block_str.insert(0, '{');
+        block_str.push('}');
+        let block = syn::LitStr::new(&block_str, s.span()).parse::<Block>()?;
+        Ok(Self(block))
+    }
+}
+
+impl From<syn::Expr> for BlockContents {
```
TedDriggs, line 81: Is this needed outside of testing?

TedDriggs: No, but the tests that use it are far enough apart as to be annoying to have this anywhere else.

Issue #232, opened 2022-02-28
Rework reference-style documentation


TedDriggs:
We have two parallel problems:

1. It's hard for crate users to find all the options for builder attributes and their effects
2. It's hard for crate maintainers to make sure new options are findable and documented

These feed on each other: As primary maintainer, _I_ don't know where best to document #228, and therefore no crate user will know where to read about how to control the creation of constructors or build functions.

We can separate documentation into some areas:

1. High-level introduction
2. Use-case examples
3. Attribute-level references

This issue is going to focus on item #3: Reference-style documentation.

---

Thanks to `darling`, our attributes are actually fairly understandable in `darling_opts.rs` - it's easy to see the root options item, and then each of the fields from there, and to recurse down to see what each one does. Unfortunately, those aren't exposed by rustdoc because those structs aren't public.

Question 1: How best to surface these? Big reference in `README.md`? Crate-level docs in the readme block?

PR #225, opened 2022-02-15
Document that the builder struct name can be controlled

ijackson:
I looked for this feature in the docs.  Eventually I found a mention of it in #157.

I felt it would be nice for it to be mentioned in the docs (other than the changelog) too :-).

Issue #224, opened 2022-02-11
Custom setter function


rezaalavi:
I would like to be able to opt-out of the setter generation for certain fields so that I can write my own setters,

Lets say I have the following struct:
```rust
#[derive(Debug, Default,Builder )]
pub struct Email{
    username:String,
    domain: String,
    #[builder(setter(skip))]
    password: String,
    quota: u64,
```

and I want to have the following code for the password setter:
```rust
impl EmailBuilder {
    pub fn with_password(&mut self, password: String) -> &mut Self {
        self.password = Some( pwhash::bcrypt::hash(password).unwrap());
        self
    }

```

The current implementation for the "setter(skip)" changes the 'password' field type to "PhantomData" which is not usable.

PR #233, opened 2022-03-01
WIP: Create API reference

TedDriggs:
This reference lists all supported attributes alphabetically.
It should in future link to examples and higher-level documentation.

Part of #232

### DIFF
--- a/derive_builder/API-REFERENCE.md
+++ b/derive_builder/API-REFERENCE.md
@@ -6,7 +6,7 @@ Applied to the struct.
 
 ```rust
 #[derive(Builder)]
-#[builder(OPTIONS HERE)]
+#[builder(OPTIONS)]
 pub struct Example {
     // ... fields elided
 }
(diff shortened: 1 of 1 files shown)


TedDriggs:
@andy128k can you take a look at this? I'm going to keep working away at getting all the options captured, and would appreciate a second set of eyes looking for errors and omissions.

andy128k:
I think it would be hard to maintain such document unless it is managed/covered by `cargo doc`.

TedDriggs:
I wish there was a way to use `cargo doc` over our options structs to auto-generate this, but the options structs aren't public; `derive_builder_core` deliberately only exposes `builder_for_struct(DeriveInput) -> TokenStream` so that adding new options doesn't force us to rev the crate version.

Even `serde` doesn't seem to have a good way of generating this documentation; it's kept separately ([example](https://github.com/serde-rs/serde-rs.github.io/blob/master/_src/field-attrs.md))

The current approach with a flat list of features in README.md isn't working - see #224 and #225 as recent examples where that's failed.

I'm hoping that the rate of API changes for `derive_builder` will be slow going forward; the crate seems pretty mature, and I'm increasingly conservative in what changes we should accept. I therefore _think_ it'll be possible to manually enforce that any macro API change PR must have an update to this file.

indietyp:
There are also things like: https://docs.rs/binread/latest/binread/attribute/index.html, which create an additional documentation-only module.

PR #236, opened 2022-03-07
Document custom setter

### LINKED ISSUE #224: Custom setter function

ijackson:
Fixes #224


derive_builder/src/lib.rs
```diff
@@ -178,6 +178,34 @@
...
+//! # #[macro_use]
+//! # extern crate derive_builder;
+//! #
+//! #[derive(Builder)]
+//! struct SetterOptOut {
+//!     #[builder(setter(custom))]
+//!     custom_setter: u32,
+//! }
+//! impl SetterOptOutBuilder {
+//!     fn set_custom_setter(&mut self, value: u32) {
```
TedDriggs, line 197: Can you have this be named `custom_setter` to show that there is no generated setter this collides with?

derive_builder/src/lib.rs
```diff
@@ -178,6 +178,34 @@
...
+//!         self.custom_setter = Some(value);
+//!     }
+//! }
+//! # fn main() {}
+//! ```
+//!
+//! Again, the more verbose form is accepted:
+//!
+//! - `#[builder(setter(custom = "true"))]`
```
TedDriggs, line 206: I don't think quotes are needed here.

TedDriggs (changes requested):
Thanks for doing this! Minor changes requested, but then good to go.

derive_builder/src/lib.rs
ijackson: Ah yes.  I got that from the `skip` example.  I've removed the quotes from both and amended one of the tests to exercise this case.

derive_builder/src/lib.rs
ijackson: Done, thanks.

Issue #221, opened 2021-11-19
Adding #[wasm_bindgen] to Builder


thorlucas:
Hi,

Is it possible to add this macro to the builder? Maybe just behind a feature flag.

Thanks,

TedDriggs:
Is it necessary in that situation to add it to the setter methods as well?

thorlucas:
I believe it only needs to be implemented on any methods that must be available from JavaScript.

TedDriggs:
On the surface this sounds possible, but I have some concerns around how it will play with other features:

* Will it work with `&mut self` setters?
* Will it work with `#[builder(setter(into))]` setters, which have type parameters in their arguments?
* Will it work with the new `each` setters?

It's important to me that new features largely interoperate with other existing features; documentation of proc-macros is still difficult, so having too many edge cases would make the crate very hard to use.

I currently have other time commitments that prevent me from working on this, but if someone else wants to take on the implementation and testing I'll happily review the results.

TedDriggs:
I don't see good answers to the above questions on the horizon, and don't think `derive_builder` can take on the test burden of making this fully supported.

That said, I also don't see as many `Builder`-style APIs in JS; much more often, I see APIs that take an `Options` plain JS object and use that to initialize the class. I might suggest using the builder for construction in Rust, but then having another function that is `#[wasm_bindgen]` that takes a `JsObject` and initializes your type from that.

One could even make a proc-macro for it, e.g. `#[derive(OptionsConstructor)]` - that would be able to handle the parsing of the object and mapping of its properties to fields on the target struct.

TedDriggs:
This is now actually possible using `#[builder_setter_attr(wasm_bindgen)]`, but you'll have to be careful not to use features that break WASM (e.g. generics, `&mut self`, etc.) - `derive_builder` will not do any testing of compatibility between the options you've chosen for your builder and additional attributes you choose to apply.

PR #237, opened 2022-03-07
Allow specifying arbitrary attributes on the builder struct

ijackson:
In arti, we would like to be able to make the builder `Deserialize`.  That means that we need to specify our custom serde attributes.  There may be other use cases.


TedDriggs:
I am very conflicted about this PR.

I agree that this is powerful, and that #221 - and probably others - would be able to benefit from it. At the same time, I have two specific fears:

1. Only attributes that parse to `syn::Meta` will work here - any other attributes will break the entire crate.
2. Does this push us too much into the realm of macros obscuring intent? I don't envy the person who has to try and debug a serde impl they can't see.

Can you elaborate a bit on why you need to deserialize into a builder? That might help me get over fear 2.

ijackson:
Hi.

Thanks for taking a look.  I agree that this is powerful and I can see why you want to be careful.  I guess my primary response is that I think whether this obscures things too much, or is useful, is a tradeoff that ought to be considered primarily by the project who choose to *use* this feature (or, indeed, choose *not* to use it).

> Can you elaborate a bit on why you need to deserialize into a builder? That might help me get over fear 2

Of course.

We want this in [Arti](https://tpo.pages.torproject.net/core/arti/) for the configuration.  Making the builder `Deserialize` will allow us to provide isomorphic configuration settings via a configuration file reader, as we do via builders and setter methods.

Currently we attempt to do this by having the *built* struct be `Deserialize`.  But this has a number of problems.  All the defaults, field renaming, and so on, have to be specified twice.  We would also like the configuration (as read from a file) to be `Serialize` as well, but there are serious problems with trying to round-trip: we would have to undo any compatibility fixups, which (in the general case) is not possible.

> Only attributes that parse to `syn::Meta` will work here - any other attributes will break the entire crate.

I think you are right and this is wrong.  It ought to be capable of handling anything that fits into the content of a `syn::Attribute`.  Should I change this so that one writes `#[builder(attrs( #[serde(whatever)] ))` ?

ijackson:
@TedDriggs Thanks for the further review and the very helpful comments.  I agree that 3 is the best approach.  I will rework this MR accordingly.

ijackson:
I tried the `darling(and_then...)` but the attributes need to be processed separately for fields, compared to setter methods, etc.  I want to avoid making a copy of the attribute list when this feature isn't used, so the manipulated attribute list wants to be `Cow`.  That means putting it in `Field` is awkward, since it would make `Field` self-referential.

Having played about, I think the best way to do this is to pass an error vec to `Field::as_builder_field`, so that it can become "fallible" (or, strictly, have an ability to report errors), and collect the errors in `builder_for_struct`.

ijackson:
> `and_then` should be taking ownership of the whole `Field`, so you're not copying anything; you'll do an in-place mutation of the array to swap out `#[builder_field_attr...]` attributes for their unnested equivalents in-place.

Well, yes, but `Field` would want to contain:
```
  attrs: Vec<Attribute>,
  field_attrs: Cow<&[Attribute]>,
```
where if the nested attribute feature isn't used, `field_attrs = Cow::Borrowed(attrs)`.  But that's not possible because that would make `Field` a self-referential struct.  This occurs in *any* situation where this "perhaps a separate value, but perhaps a reference to some other part of the options" is contained within a single Rust value along with the other options, which is necessarily the case if it's all just the return value from `from_derive_input`.

TBH, looking at the existing code, my first thought of where to add the necessary additional processing was in the construction of `FieldWithDefaults`.  I(t could contain a `Cow` referencing parts of `Field`.  But `FieldWithoutDefaults` is constructed late and infallibly.  Making its construction fallible (or at least, giving it access to `&mut errs`) would solve that but, without other changes, would involve errors arising later.

The alternatives would seem to be:

 * Never do this reference trick, and instead always copy things even when perhaps it's not needed.

 * Use `Rc`.

 * Do this later in the construction of `FieldWithDefaults` or even later in one of its methods.

 * Something like
```
    let opts = macro_options::RawOptions::from_derive_input(&ast)?; // from darling
    let opts: DefaultedAndProcessedOptions<'_> = opts.resolve()?; // can reference RawOptions
    // presumably `DefaultedAndProcessedOptions` contains `Vec<FieldWithDefaults<'a>>`.
```

Given what you've just said, I think possibly `DefaultedAndProcessedOptions` would be the right way to go about it?  That's a bit similar to what you have already done with `FieldWithDefaults`.

Anyway, I'm happy to do it whatever way you prefer so please let me know.  Thanks for your attention.

derive_builder/tests/compile-fail/builder_setter_attrs.rs
```diff
@@ -0,0 +1,19 @@
...
+#[macro_use]
+extern crate derive_builder;
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+pub struct Lorem {
+    ok: String,
+
+    // This foolish repr() attribute generates an unused attribute warning
+    #[builder_setter_attrs(
+        #[must_use]
```
TedDriggs, line 12: How does rustfmt feel about these?

TedDriggs (approved):
I hadn't anticipated putting multiple attributes inside the outer container like this, but as long as rustfmt can handle it then I'm content with the approach. Thanks for digging so deeply into this.

derive_builder/tests/compile-fail/builder_setter_attrs.rs
ijackson: Well, it doesn't *mind* them, in the sense that it is happy to process and reformat other things in the file, including the decorated field name and type.  But rustfmt doesn't seem to want to modify the nested attribute itself.  I tried misindenting the `#[must_use]` in the test case, and rustfmt didn't fix it.  It *does* still fix the indentation of the actual field name and type.

There would be no real difficulty with accepting this syntax instead:
```
    #[builder_setter_attrs(must_use)]
```
I experimented with this and then rustfmt *did* see it and re-indent it.

I chose the syntax you see not because it was easier to implement (although it was, a bit) but because I liked the way that it makes the syntax and semantics of the contained attributes very plain.  The reader doesn't have to ask themselves questions about `#[` etc.  (That each use of `builder_setter_attrs` might contain zero or more attributes follows from the syntax; it seemed pernickety to insist on precisely one, or to insist on `builder_setter_attrs` not being repeated.)

OTOH [`cfg_attr`](https://doc.rust-lang.org/reference/conditional-compilation.html#the-cfg_attr-attribute) provides precdent for doing it the other way.  Personally I find the syntax of `cfg_attr` unnecessarily confusing.

Up to you :-).  (I think you will agree with me that we should pick *one* of these syntaxes, rather than accepting either, even though the latter might be possible in theory.)

TedDriggs: I certainly agree with only accepting one syntax, and I am warming to the approach you've taken. I don't find `cfg_attr` that confusing, but I could see having one `#[builder_field_attrs` being nicer than multiple for `#[builder_field_attr(...)]` declarations.

I'm going to think about this during the day today.

ijackson: Cool, thanks.  I had another thought, which is that `#[builder_field_attrs( #[serde(with...)] )]` has the advantage that it is more likely to show up in a narrowly-targeted grep.

Also, please let me know whether you'd prefer a merge, or a rebase, for (i) if you want me to change the syntax, and (ii) to deal with the textual merge conflict with #238.

TedDriggs (changes requested):
Per comment thread, let's switch to `#[builder_field_attr(serde(...))]`.

I briefly considered the following alternate syntax:

```
#[builder_attr(field, serde(...))]
```

However, I think the single ident for the attr is sufficient, and I didn't like the risk of rustfmt putting `field` on its own line in that case.

ijackson:
OK, thanks.  Will get back to you today.

ijackson:
OK, made the syntax change.  I chose to put it on as a commit on top, rather than squash it in.

I have rebased the branch onto current master.

derive_builder/tests/compile-fail/builder_field_attrs.rs
```diff
@@ -0,0 +1,13 @@
+#[macro_use]
+extern crate derive_builder;
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+pub struct Lorem {
+    ok: String,
+
+    // This foolish repr() attribute generates an unused attribute warning
```
TedDriggs, line 8: What's `repr` referring to here?

derive_builder/tests/compile-fail/builder_setter_attrs.rs
```diff
@@ -0,0 +1,17 @@
+#![deny(unused_must_use)]
+
+#[macro_use]
+extern crate derive_builder;
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+pub struct Lorem {
+    ok: String,
+
+    // This foolish repr() attribute generates an unused attribute warning
```
TedDriggs, line 10: What's `repr` referring to here?

derive_builder/src/lib.rs
```diff
@@ -565,9 +565,7 @@
 //! #[derive(Builder)]
 //! #[builder(derive(serde::Serialize))]
 //! struct Lorem {
-//!     #[builder_field_attrs(
-//!         #[serde(rename="dolor")]
-//!     )]
+//!     #[builder_field_attrs(serde(rename="dolor"))]
```
TedDriggs, line 568: We should only be allowing one attr here, for consistency with `cfg_attr`.

TedDriggs (changes requested):
Getting closer - I don't think we want to allow multiple attributes in a single `#[builder_field_attr(...)]` attribute under this syntax. Limiting it to one makes it more like `cfg_attr`, and I'm skeptical it adds much verbosity - hopefully people aren't adding that many distinct proc-macros to their builders.

derive_builder/tests/compile-fail/builder_setter_attrs.rs
ijackson: This was leftover from an earlier version of this test case.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -241,6 +247,80 @@ pub struct Field {
...
+        Ok(self)
+    }
+}
+
+fn unnest_from_one_attribute(attr: syn::Attribute) -> darling::Result<Vec<Attribute>> {
+    match &attr.style {
+        syn::AttrStyle::Outer => (),
+        syn::AttrStyle::Inner(bang) => {
+            return Err(darling::Error::unsupported_format(
+                "builder_field_attrs must be an outer attributes",
```
TedDriggs, line 296: Nit: This function is called for `builder_field_attr` and `builder_setter_attr`; maybe get the path? Not a big deal - I'm not even sure how you'd legitimately get an inner attribute in this situation.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -241,6 +247,80 @@ pub struct Field {
...
+
+impl Field {
+    /// Remove the `builder_field_attrs(...)` packaging around attributes meant for fields
+    /// in the builder.
+    fn unnest_attrs(mut self) -> darling::Result<Self> {
+        let mut errors = vec![];
+
+        for attr in self.attrs.drain(..) {
+            let unnest = {
+                if attr.path.is_ident("builder_field_attrs") {
```
TedDriggs, line 264: TODO: Replace with `builder_field_attr` (applies throughout)

ijackson: Already on it, thanks.

derive_builder_core/src/macro_options/darling_opts.rs
ijackson: Good catch on the message, but I think as you say this error can't happen.  I have made it more general in the text, rather than including the path, and added an in-code comment.  (Will push shortly.)

ijackson:
Pushed.  Again, I didn't squash anything, not even fixup for the botched merge which I just spotted.  Every commit ought to pass the tests still so it shouldn't break bisectability...

TedDriggs (approved):
Beautiful work.

TedDriggs:
Between this and the other changes you've put in, we have enough to merit a publication. Version 0.11.0 will be available once CI passes.

ijackson:
Thanks :-).

PR #243, opened 2022-03-17
Break out distribute_and_unnest_attrs

ijackson:
RFC, as discussed in https://github.com/colin-kiegel/rust-derive-builder/pull/241#issuecomment-1069346362   .  It seemed better to demonstrate it rather than talk in the abstract.

I think it has come out quite nicely.  In particular, removing the duplicated specification of the output field variables now means it's not possible to accidentally get the unnesting and copying information out of step.

But if you prefer the code the way it is, please do feel free to say so and just close this MR.

### DIFF
--- a/derive_builder_core/src/macro_options/darling_opts.rs
+++ b/derive_builder_core/src/macro_options/darling_opts.rs
@@ -254,37 +254,52 @@ pub struct Field {
 }
 
 impl Field {
-    /// Remove the `builder_field_attr(...)` packaging around an attribute
+    /// Populate `self.field_attrs` and `self.setter_attrs` by draining `self.attrs`
     fn unnest_attrs(mut self) -> darling::Result<Self> {
-        let mut errors = vec![];
-
-        for attr in self.attrs.drain(..) {
-            let unnest = {
-                if attr.path.is_ident("builder_field_attr") {
-                    Some(&mut self.field_attrs)
-                } else if attr.path.is_ident("builder_setter_attr") {
-                    Some(&mut self.setter_attrs)
-                } else {
-                    None
-                }
-            };
-            if let Some(unnest) = unnest {
-                match unnest_from_one_attribute(attr) {
-                    Ok(n) => unnest.push(n),
-                    Err(e) => errors.push(e),
-                }
-            } else {
-                self.field_attrs.push(attr.clone());
-                self.setter_attrs.push(attr);
-            }
-        }
-
-        if !errors.is_empty() {
-            return Err(darling::Error::multiple(errors));
-        }
-
+        distribute_and_unnest_attrs(
+            &mut self.attrs,
+            &mut [
+                ("builder_field_attr", &mut self.field_attrs),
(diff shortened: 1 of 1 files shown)


derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -254,37 +254,52 @@ pub struct Field {
...
+    }
+}
 
-        if !errors.is_empty() {
-            return Err(darling::Error::multiple(errors));
+/// Handles incoming attributes from darling and assigns them to decoratee attribute lists
+///
+/// `outputs` lists each pass-through attribute name, and the corresponding decoratee attribute list.
+/// Matching atttributes are unnested and applied to the specific decoratee;
+/// Other attributes are not unnested, and simply copied for each decoratee.
```
TedDriggs, line 274:
```suggestion
/// Divide a list of attributes into multiple partially-overlapping output lists.
///
/// Some attributes from the macro input will be added to the output in multiple places;
/// for example, a `cfg` attribute must be replicated to both the struct and its impl block or
/// the resulting code will not compile.
///
/// Other attributes are scoped to a specific output by their path, e.g. `builder_field_attr`.
/// These attributes will only appear in one output list, but need that outer path removed.
///
/// For performance reasons, we want to do this in one pass through the list instead of
/// first distributing and then iterating through each of the output lists.
///
/// Each item in `outputs` contains the attribute name unique to that output, and the `Vec` where all attributes for that output should be inserted.
/// Attributes whose path matches any value in `outputs` will be added only to the first matching one, and will be "unnested".
/// Other attributes are not unnested, and simply copied for each decoratee.
```

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -254,37 +254,52 @@ pub struct Field {
...
+fn distribute_and_unnest_attrs(
+    input: &mut Vec<Attribute>,
+    outputs: &mut [(&'static str, &mut Vec<Attribute>)],
+) -> darling::Result<()> {
+    let mut errors = vec![];
+
```
TedDriggs, line 280: Is it worth adding some precondition asserts here:

```rust
for (name, list) in &outputs {
    assert!(list.is_empty(), "Output Vec for '{}' was not empty", name);
}
```

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -254,37 +254,52 @@ pub struct Field {
...
+    for attr in input.drain(..) {
+        let unnest = outputs
```
TedDriggs, line 282: Maybe name this `destination`?

derive_builder_core/src/macro_options/darling_opts.rs
ijackson: Sure.

derive_builder_core/src/macro_options/darling_opts.rs
ijackson: Good idea.  Renamed both bindings.

PR #244, opened 2022-03-21
Update darling dependency

umanwizard:
This updates the dependency on `darling` to the latest published version.

Doing so will help us and maybe others avoid multiple dependencies on `darling` in our Cargo.lock (we try to avoid duplicate dependencies as much as possible, to save on build time).

It shouldn't hurt anything for anyone else, so I hope you will accept it.

### DIFF
--- a/derive_builder_core/Cargo.toml
+++ b/derive_builder_core/Cargo.toml
@@ -17,7 +17,7 @@ readme = "README.md"
 clippy = []
 
 [dependencies]
-darling = "0.12.4"
+darling = "0.13.1"
 proc-macro2 = "1.0.24"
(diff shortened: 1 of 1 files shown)


TedDriggs:
No objections, but since it's a dependency minor-version bump I'll also need to do a crate minor-version bump to 0.12.0

TedDriggs:
Update on this; planning to do a bump to darling soon; I've got a couple changes waiting to go out with it around error accumulation.

TedDriggs:
This has been superseded by #250, which moved to `darling` 0.14.0 to pick up some new features. After #245 lands, I plan to push out an update to `derive-builder`.

PR #245, opened 2022-03-24
Support custom fields in the builder struct

ijackson:
In Arti, we want our builder structs to sometimes contain particular types that we specify, other than `Option<TargetFieldType>`.  This will allow us to control what the `[De]serialize` of the builder looks like.

There are a number of other conceivable uses for this escape hatch.  The internal machinery to implement this could also be used for other feature(s).


TedDriggs:
FYI, given the size of this PR I probably won't be able to get to it until next week.

ijackson:
> FYI, given the size of this PR I probably won't be able to get to it until next week.

Sure, thanks for the update.  FYI I will be away next week, so I will hope to read your thoughts when I get back.  I hope you find my structuring into individual commits helpful.

TedDriggs (commented):
This is quality work. I'm on the fence about whether this pushes `derive_builder` too far into the realm of arbitrary code generation - or if there's a way to unbundle what `derive_builder` does such that these advanced scenarios feel like composition rather than ever-deeper modification.

derive_builder_core/src/parsed_literal.rs
```diff
@@ -0,0 +1,23 @@
...
+
+impl<T: syn::parse::Parse> darling::FromMeta for ParsedLiteral<T> {
+    fn from_value(value: &syn::Lit) -> darling::Result<Self> {
+        if let syn::Lit::Str(s) = value {
+            Ok(ParsedLiteral(s.parse()?))
+        } else {
+            Err(darling::Error::unexpected_lit_type(value))
+        }
+    }
+}
```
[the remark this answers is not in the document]
TedDriggs: That's now live as 0.13.3

ijackson:
> This is quality work. I'm on the fence about whether this pushes `derive_builder` too far into the realm of arbitrary code generation - or if there's a way to unbundle what `derive_builder` does such that these advanced scenarios feel like composition rather than ever-deeper modification.

Thanks for the review.  I need to read the details again properly to get to grips with things (and remind myself of the structure...)

I too find the shape of this escape hatch rather uncomfortable.  It's not super convenient and the syntax is awkward.  But I think it is necessary precisely to *enable* composition: allowing the user to insert arbitrary fields into the builder structure, thus letting the user compose autogenerated code from `derive_builder` with their own handling of special cases.  The new plumbing in `derive_builder` is close to the minimum necessary to make that work.

I considered whether it woudl be possible to allow adding fields to the builder without any corresponding field in the output structure; but that would involve smuggling the whole definition of a field through magic attributes.  Or providing an attribute macro, which can of course filter the input arbitrarily.  I didn't like either of those ideas.

derive_builder_core/src/builder_field.rs
```diff
@@ -70,6 +96,17 @@ impl<'a> ToTokens for BuilderField<'a> {
...
+    fn to_tokens(&self, tokens: &mut TokenStream) {
+        match self {
+            BuilderFieldType::Optional(ty) => tokens.append_all(quote!(
+                ::derive_builder::export::core::option::Option<#ty>
+            )),
+            BuilderFieldType::Precisely(ty) => ty.to_tokens(tokens),
+        }
+    }
+}
+
```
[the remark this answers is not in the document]
ijackson: Yes, done.

derive_builder_core/src/builder_field.rs
```diff
@@ -86,7 +123,7 @@ macro_rules! default_builder_field {
     () => {{
         BuilderField {
             field_ident: &syn::Ident::new("foo", ::proc_macro2::Span::call_site()),
-            field_type: &parse_quote!(String),
+            field_type: BuilderFieldType::Optional(Box::leak(Box::new(parse_quote!(String)))),
```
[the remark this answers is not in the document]
ijackson: This is only used for tests, where this seems fine to me.  For now I have added `#[cfg(test)]`.

If you don't want this even in test  code, I could change it to `Cow`.  That would be a (very minor) perf regression for production code, existing to support only tests - which is why I felt it better to have the tests leak.

An alternative would be to change `default_builder_field` to something like `with_default_builder_field`.  That would make it possible to make it a function.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -253,9 +256,45 @@ pub struct Field {
...
+        {
+            let m = || {
+                darling::Error::unsupported_format(
+                "Specified both #[builder(default=)] and #[builder(custom(build=))] on same field, but this is not meaningful",
+            )
+            };
+            return Err(darling::Error::multiple(vec![
+                m().with_span(default_spec),
+                m().with_span(build_spec),
+            ]));
```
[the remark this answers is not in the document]
ijackson: OK.  Personally I think it's better to produce a completely duplicated diagnostic than to point to only one of the trouble causes, but they are generally going to be right next to each other.  But, it's not a big deal. I have changed this in 3532e5a5beb10b875c22ad72d99cbe92b715f8ed .

derive_builder_core/src/parsed_literal.rs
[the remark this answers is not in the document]
ijackson: Cool, thanks ;-).

derive_builder_core/src/builder_field.rs
```diff
@@ -47,20 +46,47 @@ pub struct BuilderField<'a> {
     pub attrs: &'a [syn::Attribute],
 }
 
+/// The type of a field in the builder struct, implementing `quote::ToTokens`
+#[derive(Debug, Clone)]
+pub enum BuilderFieldType<'a> {
+    /// The corresonding builder field will be `Option<field_type>`.
+    Optional(&'a syn::Type),
+    /// The corresponding builder field will be just this type
+    Precisely(&'a syn::Type),
```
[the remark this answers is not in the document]
ijackson: I have renamed this to `Precise`.

derive_builder_core/src/builder_field.rs
```diff
@@ -47,20 +46,47 @@ pub struct BuilderField<'a> {
     pub attrs: &'a [syn::Attribute],
 }
 
+/// The type of a field in the builder struct, implementing `quote::ToTokens`
+#[derive(Debug, Clone)]
+pub enum BuilderFieldType<'a> {
```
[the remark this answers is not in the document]
ijackson: Yes, this was rather strange.  I looked at it again, and played around a bit.

What I came up with was to fold `field_enabled` into a new variant `BuilderFieldType::Phantom`.  It seems more principled now.

derive_builder_core/src/initializer.rs
```diff
@@ -59,28 +59,55 @@ pub struct Initializer<'a> {
...
-            let default = self.default();
-            tokens.append_all(quote!(
-                #struct_field: #default,
+                }
             ));
         }
+
+        tokens.append_all(quote!(
+            ,
+        ));
```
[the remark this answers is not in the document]
ijackson: Mmm, yes.  It was easy to move the rhs value constrution into a closure, eliminating the possibility of such a bug.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -245,6 +245,9 @@ pub struct Field {
     default: Option<DefaultExpression>,
     #[darling(default)]
     try_setter: Flag,
+    /// Custom builder field type and build method
+    #[darling(default)]
+    custom: Option<CustomField>,
```
TedDriggs, line 250: Could this go in `field` instead?

```rust
#[builder(field(private, type = "...", build = "..."))]
```

That feels pretty nice.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -245,6 +245,9 @@ pub struct Field {
     default: Option<DefaultExpression>,
     #[darling(default)]
     try_setter: Flag,
+    /// Custom builder field type and build method
+    #[darling(default)]
```
TedDriggs, line 249: This is no longer needed; `darling` will automatically handle it because the field is an `Option`.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -253,9 +256,39 @@ pub struct Field {
     setter_attrs: Vec<Attribute>,
 }
 
+/// Options for the field-level `custom` property
+#[derive(Debug, Clone, FromMeta)]
+pub struct CustomField {
+    #[darling(rename = "type")]
+    ty: syn::Type,
+    #[darling(default)]
```
TedDriggs, line 264: This is no longer needed; `darling` will automatically make it optional.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -253,9 +256,39 @@ pub struct Field {
...
-    /// Populate `self.field_attrs` and `self.setter_attrs` by draining `self.attrs`
-    fn unnest_attrs(mut self) -> darling::Result<Self> {
+    /// Resolve and check (post-parsing) options which come from multiple darling options
+    ///
+    ///  * Check that we don't have a custom field builder *and* a default value
+    ///  * Populate `self.field_attrs` and `self.setter_attrs` by draining `self.attrs`
+    fn resolve(mut self) -> darling::Result<Self> {
+        // Perhaps at some future point there would be a way to specify the default value of the field in the builder.
+        // That ought possibly to be distinct from the default value of the field in the target.
+        // For now, reject this situation.
```
TedDriggs, line 276: I don't see us ever doing that.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -686,11 +719,32 @@ impl<'a> FieldWithDefaults<'a> {
     }
 
     pub fn field_vis(&self) -> Visibility {
-        self.field
-            .field
-            .as_expressed_vis()
-            .or_else(|| self.parent.field.as_expressed_vis())
-            .unwrap_or(Visibility::Inherited)
+        if !self.field_enabled() {
```
TedDriggs, line 722: Why does this now preempt the `#[builder(field(...))]` option?

derive_builder_core/src/setter.rs
```diff
@@ -210,6 +223,11 @@ impl<'a> ToTokens for Setter<'a> {
     }
 }
 
+/// Returns expression wrapping `bare_value` in `Some`
+pub fn wrap_expression_in_some(bare_value: TokenStream) -> TokenStream {
```
TedDriggs, line 227: Does this need to be `pub`?

derive_builder_core/src/builder_field.rs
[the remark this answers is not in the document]
TedDriggs: I agree this version is more principled. I am now wondering why the original author felt the need to use `PhantomData` in the case a setter was skipped; is saving the size of a skipped field worth the added complexity, particularly if someone could reclaim the bytes using this new custom field type?

Would excising this phantom completely be a significant complexity win (and could someone use PhantomData themselves as a custom field type if they wanted?)

derive_builder_core/src/macro_options/darling_opts.rs
ijackson: Ah yes.

derive_builder_core/src/setter.rs
ijackson: Oh, no, not any longer.

derive_builder_core/src/macro_options/darling_opts.rs
ijackson: Would you like to suggest a different wording for the comment?

derive_builder_core/src/macro_options/darling_opts.rs
ijackson: Yes, it turns out.

derive_builder_core/src/builder_field.rs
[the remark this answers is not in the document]
ijackson: Personally I think the PhantomData is worth the complexity.  I have added a justification as a comment :-).

derive_builder_core/src/macro_options/darling_opts.rs
ijackson: I have added a comment about this.  This preserves the behaviour from before my MR.  Previously this private visibility was dealt with in ad-hoc code in the builder struct code generator.  Now, the field *type* comes from `BuilderFieldType`, and the code generator always uses the visibility which comes from `field_vis`, so `field_vis` must return the desired answer.

ijackson:
I am going to rebase this onto master now, to resolve the conflict.  I won't make other changes as part of the rebase.

derive_builder_core/src/macro_options/darling_opts.rs
TedDriggs: First: Great comment.

Second: I agree with removing the ad-hoc hiding in codegen, so let's keep the logic for visibility all here.

Third: With #247, I feel like explicit field-level visibility should overrule the inferred visibility from being a phantom. How would you feel about this checking `self.field.field.as_expressed_vis()`, then checking if the field is enabled, then checking the struct-level? Or should it check those first with the normal inheritance chain and finally fall back to inferring `Inherited` vis if nothing is pushing to make it public?

derive_builder_core/src/setter.rs
```diff
@@ -146,14 +154,19 @@ impl<'a> ToTokens for Setter<'a> {
                     quote!(<VALUE: ::derive_builder::export::core::convert::TryInto<#ty>>);
                 let try_ident = syn::Ident::new(&format!("try_{}", ident), Span::call_site());
 
+                let mut converted = quote! {converted};
+                if builder_field_is_option {
+                    converted = wrap_expression_in_some(converted);
+                }
```
TedDriggs, line 160: Can this be rewritten as `let converted = if builder_field_is_option { ... } else { ... }`?

derive_builder_core/src/setter.rs
```diff
@@ -210,6 +223,11 @@ impl<'a> ToTokens for Setter<'a> {
     }
 }
 
+/// Returns expression wrapping `bare_value` in `Some`
+fn wrap_expression_in_some(bare_value: TokenStream) -> TokenStream {
```
TedDriggs, line 227:
```suggestion
fn wrap_expression_in_some(bare_value: impl ToTokens) -> TokenStream {
```

This would avoid forcing the caller to allocate a token stream unnecessarily.

TedDriggs (changes requested):
Marking "Request changes" to put `type` and `build` into `FieldMeta`

ijackson:
Hi.  Just got back to this after the long weekend here for Easter.  Thanks for your comments, which I'll work on now.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -237,6 +237,11 @@ pub struct Field {
     /// This property only captures the first two, the third is computed in `FieldWithDefaults`.
     default: Option<DefaultExpression>,
     try_setter: Flag,
+    /// Custom builder field type
+    #[darling(rename = "type")]
+    builder_type: Option<syn::Type>,
+    /// Custom builder field method, for making target struct field value
+    build: Option<BlockContents>,
```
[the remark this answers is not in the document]
ijackson: Sorry, I must not have read your comment properly.  I will change it.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -246,8 +251,26 @@ pub struct Field {
...
+        if let Field {
+            default: Some(_),
+            build: Some(custom_build),
+            ..
+        } = &self
+        {
+            return Err(darling::Error::unsupported_format(
+                r#"#[builder(default)] and #[builder(build="...")] cannot be used together"#,
+            )
+            .with_span(custom_build));
```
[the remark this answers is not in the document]
ijackson: It does.  I made this change.  You can see the change to the error output in one of the test cases, here: https://github.com/colin-kiegel/rust-derive-builder/pull/245/commits/fdad186c220138d37ad7a16e503073e1aa3363ef#diff-0e03ed409b7d18d2a2ce3c84031924515cc762b82018455d3d3e0c81a84e86ab

derive_builder_core/src/setter.rs
ijackson: Well, it could be
```
    let converted = if builder_field_is_option {
        wrap_expression_in_some(quote! {converted})
    } else {
        quote! {converted}
    };
```
but that has `quote! {converted}` twice.  Or
```
    let converted = quote! {converted};
    let converted = builder_field_is_option {
        wrap_expression_in_some(converted)
    } else {
        converted
    };
```
which seems hardly better, if at all.

In fact *every* call to `wrap_expression_in_some` is conditional and uses the pattern with `mut`: here, and (earlier) the two successive calls on `into_value`.

How about
```
    if (stripped_option) { wrap_expression_in_some(&mut into_value); }
...
    if (builder_field_is_option) { wrap_expression_in_some(&mut converted); }
```

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -92,6 +92,11 @@ impl FlagVisibility for BuildFn {
 pub struct FieldMeta {
     public: Flag,
     private: Flag,
+    /// Custom builder field type
+    #[darling(rename = "type")]
+    builder_type: Option<syn::Type>,
+    /// Custom builder field method, for making target struct field value
+    build: Option<BlockContents>,
```
TedDriggs, line 99: I had forgotten that `FieldMeta` is currently used by both `Field` and `Options` - these settings definitely don't make sense in the latter context, so we'll need to follow the example of `StructLevelSetter` and `FieldLevelSetter` and make two structs. The struct-level one should only have the visibility settings, while the field level one would have the visibility settings and these new fields.

```rust
/// Contents of the `field` meta in `builder` attributes.
#[derive(Debug, Clone, Default, FromMeta)]
pub struct StructLevelFieldMeta {
    public: Flag,
    private: Flag,
    vis: Option<syn::Visibility>,
}

impl Visibility for StructLevelFieldMeta {
    fn public(&self) -> &Flag {
        &self.public
    }

    fn private(&self) -> &Flag {
        &self.private
    }

    fn explicit(&self) -> Option<&syn::Visibility> {
        self.vis.as_ref()
    }
}

#[derive(Debug, Clone, Default, FromMeta)]
pub struct StructLevelFieldMeta {
    public: Flag,
    private: Flag,
    vis: Option<syn::Visibility>,
    /// Custom builder field type
    #[darling(rename = "type")]
    builder_type: Option<syn::Type>,
    /// Custom builder field method, for making target struct field value
    build: Option<BlockContents>,
}

impl Visibility for FieldLevelFieldMeta {
    fn public(&self) -> &Flag {
        &self.public
    }

    fn private(&self) -> &Flag {
        &self.private
    }

    fn explicit(&self) -> Option<&syn::Visibility> {
        self.vis.as_ref()
    }
}
```

derive_builder_core/src/macro_options/darling_opts.rs
ijackson: > I feel like explicit field-level visibility should overrule the inferred visibility

Good point.  I agree completely. about explicitly-specified field-level visibility,  But, I think, only field-level, not struct-level visibility.

derive_builder_core/src/macro_options/darling_opts.rs
ijackson: Oh.  I think the struct-level ones will always be a subset of the field-level ones, with the same names, surely?  Anything else would be very confusing.  Unfortunately darling doesn't have `#[darling(flatten)]`.

TedDriggs: Yes, but reusing the same struct would cause darling’s “did you mean” suggestions to include fields that aren’t valid at the struct level. The code duplication is annoying, but at least it’s fairly small.

ijackson: I think at this point I need to rebase onto #247, before doing as you suggest.  Is #247 likely to change significantly, or can I do that now?

TedDriggs: I've just merged it (and the last push was just a commit message edit, so it hasn't changed at all).

ijackson: Or, maybe, merge with #247.

ijackson:
Rebased onto #245, sorted out textual conflicts, with a final commit at the end to fix two semantic conflicts.

derive_builder/tests/builder_field_custom.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use std::num::ParseIntError;
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+pub struct Lorem {
+    #[builder(field(type = "Option<usize>", build = "self.ipsum.unwrap_or(42) + 1"))]
+    ipsum: usize,
+
+    #[builder(field(type = "String", build = "self.dolor.parse()?"))]
+    dolor: u32,
```
TedDriggs, line 14: Love this example; it's really clean and easy to follow.

derive_builder/tests/builder_field_custom.rs
```diff
@@ -0,0 +1,51 @@
...
+
+impl From<ParseIntError> for LoremBuilderError {
+    fn from(e: ParseIntError) -> LoremBuilderError {
+        LoremBuilderError::ValidationError(e.to_string())
+    }
+}
+
+#[test]
+fn custom_fields() {
+    let x = LoremBuilder::default().dolor("7".into()).build().unwrap();
```
TedDriggs, line 25: Can you drop the `into` here, and instead make this an `into` setter above? I think it'd be good to show that feature working in tandem with custom field types.

derive_builder/tests/compile-fail/builder_field_custom.rs
```diff
@@ -0,0 +1,10 @@
+#[macro_use]
+extern crate derive_builder;
+
+#[derive(Debug, PartialEq, Default, Builder, Clone)]
+pub struct Lorem {
+    #[builder(default = "88", field(type = "usize", build = "self.ipsum.unwrap_or_else(42) + 1"))]
```
TedDriggs, line 6: If the field type here is `usize`, how would `unwrap_or_else` work here?

derive_builder_core/src/builder_field.rs
```diff
@@ -33,45 +33,41 @@ use syn;
...
-                #(#attrs)* #ident: ::derive_builder::export::core::marker::PhantomData<#ty>,
-            ));
+        match self {
+            BuilderFieldType::Optional(ty) => tokens.append_all(quote!(
+                ::derive_builder::export::core::option::Option<#ty>
+            )),
+            BuilderFieldType::Precise(ty) => ty.to_tokens(tokens),
+            BuilderFieldType::Phantom(ty) => tokens.append_all(quote!(
+                ::derive_builder::export::core::marker::PhantomData<#ty>
+            )),
```
TedDriggs, line 53: Please move this to live with `BuilderFieldType` and its trait impls.

derive_builder_core/src/builder_field.rs
```diff
@@ -82,20 +78,55 @@ impl<'a> BuilderField<'a> {
 
 /// Helper macro for unit tests. This is _only_ public in order to be accessible
 /// from doc-tests too.
+#[cfg(test)] // This contains a Box::leak, so is suitable only for tests
 #[doc(hidden)]
 #[macro_export]
 macro_rules! default_builder_field {
```
TedDriggs, line 84: Please move this macro down to be after the `BuilderFieldType` declaration.

derive_builder_core/src/initializer.rs
```diff
@@ -59,28 +59,47 @@ pub struct Initializer<'a> {
...
+    /// Method to use to to convert the builder's field to the target field
+    ///
+    /// For sub-builder fields, this will be `build` (or similar)
+    pub conversion: FieldConversion<'a>,
 }
 
 impl<'a> ToTokens for Initializer<'a> {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         let struct_field = &self.field_ident;
+        let builder_field = &*struct_field;
```
TedDriggs, line 71: If you can't write `let builder_field = struct_field;` (which is surprising, since I'd have guessed it impls Copy), then how about this?

```suggestion
        let builder_field = &self.field_ident;
```

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -127,13 +127,44 @@ impl Visibility for BuildFn {
 
 /// Contents of the `field` meta in `builder` attributes.
```
TedDriggs, line 128:
```suggestion
/// Contents of the `field` meta in `builder` attributes at the struct level.
```

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -127,13 +127,44 @@ impl Visibility for BuildFn {
...
+    fn explicit(&self) -> Option<&syn::Visibility> {
+        self.vis.as_ref()
+    }
+}
+
+/// Contents of the `field` meta in `builder` attributes.
+//
+// TODO this struct and its Visibility impl duplicate code from StructLevelFieldMeta
+// Ideally they would be combined somehow, but without adversely affecting diagnostics
+// from darling, etc.
```
TedDriggs, line 154:
```suggestion
/// Contents of the `field` meta in `builder` attributes at the field level.
///
/// This is a superset of the attributes permitted in `field` at the struct level.
```

The TODO is true, but it's no different than `StructLevelSetter`, which doesn't have any such comment. I've reopened TedDriggs/darling#146 to evaluate adding `flatten` there; maybe link this to that?

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -302,8 +333,30 @@ impl Field {
...
+                    build: Some(_custom_build),
+                    ..
+                },
+            ..
+        } = &self
+        {
+            return Err(darling::Error::unsupported_format(
+                r#"#[builder(default)] and #[builder(build="...")] cannot be used together"#,
+            )
+            .with_span(field_default));
```
TedDriggs, line 357: Please use `Error::accumulator()` here to avoid losing errors from `distribute_and_unnest_attrs` below.

derive_builder_core/src/setter.rs
TedDriggs: Hmm; I don't mind the `quote!(converted)` appearing twice, but I don't feel strongly enough about it to delay this PR.

TedDriggs (changes requested):
Getting close now.

derive_builder_core/src/initializer.rs
ijackson: Just removing the `&*` works. Not sure why it was there; perhaps some earlier version of this code in my branch needed it.

ijackson:
Thanks for the detailed review!  I think I have dealt with all your comments.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -147,11 +147,11 @@ impl Visibility for StructLevelFieldMeta {
...
 }
 
-/// Contents of the `field` meta in `builder` attributes.
+/// Contents of the `field` meta in `builder` attributes at the field level.
 //
-// TODO this struct and its Visibility impl duplicate code from StructLevelFieldMeta
-// Ideally they would be combined somehow, but without adversely affecting diagnostics
-// from darling, etc.
+// This is a superset of the attributes permitted in `field` at the struct level.
+// Perhaps in the future we will be able to use `#[darling(flatten0]`, but
```
TedDriggs, line 153:
```suggestion
// Perhaps in the future we will be able to use `#[darling(flatten)]`, but
```

Nit: Typo

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -351,21 +353,23 @@ impl Field {
             ..
         } = &self
         {
-            return Err(darling::Error::unsupported_format(
-                r#"#[builder(default)] and #[builder(build="...")] cannot be used together"#,
-            )
-            .with_span(field_default));
+            errors.push(
+                darling::Error::unsupported_format(
```
TedDriggs, line 357:
```suggestion
                darling::Error::custom(
```

The docs for `unsupported_format` imply it's narrowly-applicable to the case of meta-items that are lists vs name-value pairs. Here, `custom` will be better.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -810,7 +814,14 @@ impl<'a> FieldWithDefaults<'a> {
                 // Disabled fields become a PhantomData in the builder.  We make that field
                 // non-public, even if the rest of the builder is public, since this field is just
                 // there to make sure the struct's generics are properly handled.
-                || (!self.field_enabled()).then(|| Cow::Owned(syn::Visibility::Inherited)),
+                || {
+                    // We would prefer to use bool::then but that's only available in Rust 1.50, not in our MSRV
```
TedDriggs, line 818:
```suggestion
```

This is true, but the savings of switching to `then` are small enough that it's not worth the added line for the comment.

TedDriggs (approved):
All substantive feedback is addressed; there are a couple minor changes and then we'll be ready to merge.

ijackson:
> All substantive feedback is addressed; there are a couple minor changes and then we'll be ready to merge.

Cool, thanks.  Your suggestions looked good to me so I just committed them via the github UI.

TedDriggs (approved):
@ijackson what a heroic effort, and a great result.

I appreciate that you split your work out by commit, but I'd rather not add all 72 as individual commits to `master`. Would you like to do an interactive rebase locally to organize them, or would you prefer I squash and merge?

ijackson:
Thanks :-).  I think it's probably best to squash, unless you would prefer to retain a more detailed history.  If the latter, let me know.

TedDriggs:
[Squashing now](https://youtu.be/JyA1lBJl_qM?t=11)

Issue #246, opened 2022-04-06
More fine-grained field visibility


indietyp:
Hello, first off: great project!

I am currently trying to integrate the builder with my tests, and therefore I'd need the fields exposed. After looking through the code, there seems to be `#[builder(field(public))]` to switch the field to `pub,` but not to `pub(crate)` or `pub(super).`

Is there another way to set the visibility, or is this currently the only way to do things?
I'd love a struct level `[builder(field(vis = "pub(super)"))]` or something similar. Even at a field level, this would be very helpful.

indietyp:
I'd be willing to implement such a (backward-compatible) change and create a pull request if that is desired.

PR #247, opened 2022-04-06
Add restricted visibility support to builders, build methods, setters, and fields

TedDriggs:
* Use Cow&lt;syn::Visibility> to avoid unnecessary cloning
* Make as_expressed_vis return a Result
* Add Visibility::explicit

This will be updated for #246 soon.

### DIFF
--- a/derive_builder/tests/compile-fail/vis_conflict.rs
+++ b/derive_builder/tests/compile-fail/vis_conflict.rs
@@ -1,0 +1,21 @@
+#[macro_use]
+extern crate derive_builder;
+
+#[derive(Builder)]
+#[builder(public, vis = "pub(crate)")]
+pub struct Example {
+    #[builder(public, private)]
+    field: String,
+}
+
+#[derive(Builder)]
+#[builder(public, vis = "pub(crate)", build_fn(private, public))]
+pub struct SecondExample {
+    #[builder(private, vis = "pub")]
+    a_field: String,
+}
+
+fn main() {
+    ExampleBuilder::default().build();
+    SecondExampleBuilder::default().build();
+}
--- a/derive_builder_core/src/macro_options/darling_opts.rs
+++ b/derive_builder_core/src/macro_options/darling_opts.rs
@@ -25,33 +25,47 @@ trait Visibility {
 
     /// Get the explicitly-expressed visibility preference from the attribute.
     /// This returns `None` if the input didn't include either keyword.
-    fn as_expressed_vis(&self) -> darling::Result<Option<Cow<syn::Visibility>>> {
+    ///
+    /// # Panics
+    /// This method panics if the input specifies both `public` and `private`.
+    fn as_expressed_vis(&self) -> Option<Cow<syn::Visibility>> {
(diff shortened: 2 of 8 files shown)


TedDriggs:
@ijackson any idea on what's failing here? I don't repro the build failure locally on rustc 1.59.0 on macOS.

derive_builder/tests/compile-fail/vis_conflict.rs
```diff
@@ -0,0 +1,31 @@
...
+    // KNOWN ISSUE: The span for this error points to the macro invocation,
+    // rather than pointing to `public` or `private`.
+    #[builder(public, private)]
+    field: String,
+}
+
+// KNOWN ISSUE: There is no error for the struct-level conflicting visibility
+// declarations, only for build_fn. This is happening because `from_meta` for
+// `build_fn` returns `Err`, so the conflict check for the root `Options` does
+// not occur.
```
TedDriggs, line 16: I'm not thrilled with all these known issues, and it seems annoying that we refuse to emit anything in the case of ambiguous visibility (creating lots of secondary errors to deal with).

Could we instead defer conflict checking to after `FromDeriveInput::from_derive_input` has returned `Ok`, putting it in a new `check_nonfatal_errors` method, and have `as_expressed_vis` return the most conservative of the passed-in visibilities? That would mean we produce errors for all collisions, regardless of location, and that we still emit a builder struct.

TedDriggs: As an interim approach here, I've deferred the validation but left the panic in place. This makes it impossible for invalid cases to produce working code, and since I almost missed one validation path during development I think that's better than trying to continue past the issue for now.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -221,6 +264,12 @@ pub struct Field {
     pattern: Option<BuilderPattern>,
     public: Flag,
     private: Flag,
+    /// Declared visibility for the field in the builder, e.g. `#[builder(vis = "...")]`.
+    ///
+    /// This cannot be named `vis` or `darling` would put the deriving field's visibility into the
+    /// field instead.
+    #[darling(rename = "vis", default)]
```
TedDriggs, line 271: Don't need `default` here.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -425,6 +488,12 @@ pub struct Options {
 
     private: Flag,
 
+    /// Desired visibility of the builder struct.
+    ///
+    /// Do not confuse this with `Options::vis`, which is the visibility of the deriving struct.
+    #[darling(rename = "vis", default)]
```
TedDriggs, line 494: Don't need `default` here.

indietyp (approved):
looks good to me!

TedDriggs:
@ijackson can I merge this without breaking your work on custom field types too badly?

ijackson:
Please go ahead, I can fix up conflicts. I don't want to block others' work.

PR #252, opened 2022-04-19
Clarify as_expressed_vis

ijackson:
I found the old code confusing to read, since one has to pore through
it to see what the conditions for panicing are.

No functional change.

This is purely stylistic.  If you don't agree with my sense of taste here, please just close this MR :-).

Thanks.

### DIFF
--- a/derive_builder_core/src/macro_options/darling_opts.rs
+++ b/derive_builder_core/src/macro_options/darling_opts.rs
@@ -28,22 +28,23 @@ trait Visibility {
     ///
     /// # Panics
     /// This method panics if the input specifies both `public` and `private`.
-    fn as_expressed_vis(&self) -> Option<Cow<syn::Visibility>> {
-        let declares_public = self.public().is_present();
-        let declares_private = self.private().is_present();
-        let declares_explicit = self.explicit().is_some();
+    fn as_expressed_vis<'s>(&'s self) -> Option<Cow<'s, syn::Visibility>> {
+        let mut to_return = None;
+        let mut intend_return_some = |v: Cow<'s, syn::Visibility>| {
+            if let Some(already) = &to_return { panic!("visibility specified both of {:?} {:?}", already, &v); }
+            to_return = Some(v);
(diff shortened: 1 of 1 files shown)


TedDriggs:
> since one has to pore through it to see what the conditions for panicing are.

Is this due to relying on `assert` for the `panic`?

I personally prefer the current formulation, because it's closer to what I'd have written in the happy-path case where errors were impossible, and this should have been validated before `as_expressed_vis` was called. However, as the author of this function I'm not particularly qualified to decide its readability, so I am open to suggestions here.

ijackson:
The thing I found confusing was as follows: this code is trying to check that at most one of these things is specified.  But this was implemented as an ad-hoc check in each if arm that the other things weren't specified (but, only the things handled by subsequent rather than earlier branches).

The way I have rewritten it means that this property is explicit (and it's not possible to accidentally get the asserts wrong).

Another possibility would be to express the same thing by building an iterator out of the if arms, and checking that it had length 1, but that seemed abstruse.

TedDriggs:
> The thing I found confusing was as follows: this code is trying to check that at most one of these things is specified. But this was implemented as an ad-hoc check in each if arm that the other things weren't specified (but, only the things handled by subsequent rather than earlier branches).

It's worth mentioning here that I considered allowing `as_expressed_vis` to return a value without panicking, so that we could emit both the builder and the compile error about conflicting visibilities. I then got cold feet because I almost missed one case, but it would be nice to still emit the builder struct in that error case so that people don't get errors about builder structs not existing when they're trying to fix the one visibility conflict issue.

I'm comfortable with this code as-is, because:

1. I cannot imagine what else we'd add here - there's now literally support for all the visibilities a field is capable of having.
2. There are trybuild tests that would fail if the `no_visiblity_conflict` calls were missed.

I do appreciate you taking the time to review this change and provide feedback; being primary maintainer on this makes it difficult at times to have meaningful discussions about code readability, and it's great to get the chance to do so.

PR #253, opened 2022-04-21
Nested builder support

ijackson:
This builds on the codegen #245 to provide the ability to sensibly nest structs with builders.  With this, our code in Arti can do this:

```
#[derive(Clone, Builder, Debug, Eq, PartialEq, AsRef)]
#[builder(build_fn(error = "ConfigBuildError"))]
#[builder(derive(Deserialize))]
pub struct TorClientConfig {
    /// Information about the Tor network we want to connect to.
    #[builder(sub_builder)]
    tor_network: dir::NetworkConfig,
```

with the result that:

 * `TorClientConfigBuilder` contains a `NetworkConfigBuilder`.
 * Callers can do things like `tor_client_config_builder.tor_network().fallback_caches(...)`
 * There is no need for anyone to write out the `.build()` calls for the sub-structures.  `tor_client_config_builder.build()` does it.
 * `impl TorClientConfigBuilder { pub fn tor_network(&mut self) -> &mut NetworkConfigBuilder { ... } }` can now be autogenerated rather than our current handwritten methods

`sub_builder` changes defaults for the contained field type and how to construct the target field, but you can still override them with the features from #245 

I guess you will have some opinions about the shape of the field attribute syntax...

Thanks again for your attention!


TedDriggs:
Can you file an issue for this proposal?

I'd like to better understand the use-case and available alternatives before reviewing a particular implementation.

ijackson:
> I'm going to take a closer look at the usage of `sub_builder` in `arti`. 

Thanks.

One thing you might want to look at is the art MR I just made where I am trying to sort out our config API for "list-like" config fields.  Some of these are plain values, and some are themselves things with builders.

Here's the macro_rules I currently have to help drive sub_builder:

https://gitlab.torproject.org/tpo/core/arti/-/blob/06e51f63843b83506543f7055f220cb4ff0b2538/crates/tor-config/src/list_builder.rs

I think this could usefully be done by a proc macro, perhaps in derive_builder, but it has quite a large API surface.


> > I think this is a thing one would only do if there were compelling reasons why the "inner" macro couldn't be modified to add the desired features.
> 
> `serde` has to deal with this, and does so by having some low-level hooks such as `with` and then having other crates like `serde_with` or the one for location tracking of errors that callers can use to get additional functionality.

Yes.  However, of course, `serde` is already very complicated and is much more at risk of becoming an incoherent pile of random and badly-interoperating features than `derive_builder` is.

I found the fact that the proc_macro code in #253 is so small quite encouraging.

> I still have a vague sense that this is too focused on the specific sub-builder use-case,

Well, the `sub_builder` attribute that I add in my MR is of course precisely focused on that.  But it's just a shorthand for a longer set of attributes, and avoids repetition there.  I think it's appropriate for a macro package like derive_builder to have convenience affordances for common patterns.

But I guess maybe you are also looking at the new error type I introduced.  You're right that it is quite specific to the sub-builder case.  Possibly it would be better to separate out the functions of "error occurred in a particular field" and "error was a sub-builder error" - but IME flat error conversion hierarchies are normally easier to work with than repeated conversions.

So having a specific error for precisely "sub builder for field failed" doesn't seem too bad.  The logical conclusion of this argument is is that if derive_builder grew built-in support for `Option<Vec<ItemBuilder>>` like I have in my macro_rules, there would be a specific error precisely for "sub builder for entry in list in field failed", containing both the field name and the index.

> I wonder if there's a pattern around "builder attempts to initialize field within a try-block, then adds field name if error is emitted."

That sounds like a reasonable feature to me.  But there are of course *two* needless occurrences of the field name in the "just use `#[builder(field(...))]` approach: the one in the error handling, and the one for the sub-field in the outer builder (ie, the input value).

> I do have a fairly strong dislike for the fact that `field(build = "...")` embeds the field name - it's consistent with how the crate previously used blocks in attribute arguments, but this could potentially have been avoided if that took a path and called that function on the field rather than having the caller specify the field name.

`field(build = )` lets one make an output field out of multiple input fields.  ISTM that there are all sorts of reasons why that might be useful, (particularly in cases where one is trying to do some kind of backward compatibilty).  It's the fully general escape hatch which one ought only to use if absolutely necessary.

-----

Anyway, thanks for your consideration.  FYI, currently arti is using a git dependency for derive_builder.  We can't release in that state, so it is possible that we may need to upload a fork of derive_builder to crates.io.  My colleagues are not particularly worried about the implications of having a pet fork of derive_builder, but I would very much like to avoid it.

Issue #256, opened 2022-05-30
BuilderErrors derived with derive_builder do not have `PartialEq` nor `Eq` derived


EvanCarroll:
I'm wrapping the error generated by derive_buidler. The error wrapper that I'm creating derives `Eq` and `PartialEq`: it needs to derive those traits because my test suite is built by testing things that throw an error and using `assert_eq!()` to test that the error I got back is the same as the error I'm expecting. Because my wrapper error derives `Eq` and `PartialEq`, I get this error when I try to include the Error struct that derive_builder generates,

```
error[E0277]: the trait bound `ElementBuilderError: Eq` is not satisfied
   --> src/error.rs:21:3
    |
5   | #[derive(Debug, Eq, PartialEq, Error)]
    |                 -- in this derive macro expansion
...
21  |         source: crate::element::ElementBuilderError,
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `Eq` is not implemented for `ElementBuilderError`
    |
note: required by a bound in `AssertParamIsEq`
   --> /home/ecarroll/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/cmp.rs:316:31
    |
316 | pub struct AssertParamIsEq<T: Eq + ?Sized> {
    |                               ^^ required by this bound in `AssertParamIsEq`
    = note: this error originates in the derive macro `Eq` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider annotating `ElementBuilderError` with `#[derive(Eq)]`
    |
6   | #[derive(#[derive(Eq)]
    |          +++++++++++++

error[E0369]: binary operation `==` cannot be applied to type `ElementBuilderError`
  --> src/error.rs:21:3
   |
5  | #[derive(Debug, Eq, PartialEq, Error)]
   |                     --------- in this derive macro expansion
...
21 |         source: crate::element::ElementBuilderError,
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
note: an implementation of `PartialEq<_>` might be missing for `ElementBuilderError`
  --> src/element.rs:6:10
   |
6  | #[derive(Builder, Clone, Debug, PartialEq, Eq)]
   |          ^^^^^^^ must implement `PartialEq<_>`
[...]

PR #257, opened 2022-06-13
added Eq, and PartialEq

EvanCarroll:
Derive Eq and PartialEq on BuilderError. #256


EvanCarroll:
Computer crashed, lost the message I typed. I've heard of things like that before. I was unconvinced, so I wanted to do my own research on it. You may find this informative. https://stackoverflow.com/q/72609905/124486

PR #258, opened 2022-07-20
Feature/post build operation

marlon-sousa:
In this pr, we suggest a new feature, in form of the post_build attribute added to build_fn.

For functional details, please refer to #259 .

## Technical implementation

This PR implement following changes:

1. A post_build attribute is created as a valid attribute on build_fn attribute.
2. The builder function code is modified to:

    * Build the target struct and assign it to a variable called res. If a post build fn is provided, the variable is mutable, otherwise it is immutable.
    * If a post_build fn is provided, call it passing the &mut reference to the target struct.
    * Convert errors from the post build function to a instance of the PostBuildError enumeration
    * If post_build fn succeeds *** or if no post_build fn is provided, return ok(res), where res is the variable holding the built struct.
3. A PostBuildError type is provided
4. Generate, for the auto generated error, a conversion from PostBuildError to the PostBuildError variant of the auto generated error.
5. Provide documentation and examples.

For reviewers: please bear in mind that this is my first contribution to a macro crate, so I might have missed something. I know for sure that the compile fail test is not right, as I had no way of running it *** I tried to look for instruction on the documentation but I did not find anything.

Special care was taken not to cause api breaking changes: only if one uses build_fn and custom errors together should them change the impls on the customer error to include a From<PostBuildError> implementation.

### DIFF
--- a/derive_builder/examples/post_build.rs
+++ b/derive_builder/examples/post_build.rs
@@ -1,0 +1,124 @@
+// some times, one or more fields of your struct need to be built according to other field's values
+// may be you have a layout which needs tpo know all other field's values to be built, or that you have some field that needs to be calculated only once for the struct
+// and you don't want to recalculate it every time one of the fields are set in the builder
+// There are several situations in which a post build event might be useful.
+// 1. You might have a field which needs to be calculated when the struct is built, but deppends on several other field values to calculate.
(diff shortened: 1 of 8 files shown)


TedDriggs:
Per [this comment](https://github.com/colin-kiegel/rust-derive-builder/issues/259#issuecomment-1196076619), this additional option solves a problem for which this crate already has a solution.

Issue #223, opened 2022-02-06
Generate `builder` method on the target struct


stoically:
I'm almost always adding a `builder` method to the target struct, like e.g.

```rust
#[derive(Debug, Builder)]
struct Something {
  field: String,
}

impl Something {
  pub fn builder() -> SomethingBuilder {
    Default::default()
  }
}
```

Would be useful to have derive-builder generate it for me.

TedDriggs:
This is discussed [here](https://github.com/colin-kiegel/rust-derive-builder/issues/170#issuecomment-665087279); given the relatively small savings, and the impact on how the struct will present in documentation, I’m disinclined to add it.

stoically:
Thanks for the pointer. I guess that makes sense. Tho the idea to have an attribute for `#[builder()]` might be nice. Maybe something like `builder_fn`.

Issue #170, opened 2020-07-28
Create Builder from instance ( to_builder() )


mseravalli:
Add the possibility of creating a builder from an instance with `to_builder()` so that we can create new instances from an existing one.

```rust
#[derive(Builder)]
struct Lorem {
    ipsum: i32,
    dolor: i32,
}

fn do_smth() ... {
    ...
    let lorem = LoremBuilder::default().ipsum(42).dolor(0).build()?;
    let more_dolor = lorem.to_builder().dolor(1).build()?;
    ...
}
```

TedDriggs:
This is something I frequently add to my own structs, but I usually just do it with my own `impl` block since the code required is pretty short:

```rust
pub fn builder() -> LoremBuilder {
    LoremBuilder::default()
}
```

We did previously agree that we'd add it as `builder` if we added the inherent method for the caller.

# Points For
1. It's a very common ergonomic choice for calling crate offers to provide this.

# Points Against
1. If the crate auto-generates this, it will be in its own `impl` block. That shows up in the generated documentation, which some people don't like.
2. This code basically never changes: Unlike many of the builder internals, this stays the same when the fields of the builder change. That makes the time savings and correctness wins of the code generation smaller.

If we do add it, it'll need to be opt-in. What would you call in the `#[builder(...)]` attribute?

mseravalli:
Thanks for the fast answer, but, if I understand it correctly, the proposed solution does not copy the current existing values of the instance, right?
I was thinking about a solution for easily duplicating an instance and maybe overwrite a single attribute. Something like

```rust
struct Lorem {
    ipsum: i32,
    dolor: i32,
}

pub fn builder(&self) -> LoremBuilder {
    LoremBuilder::default().ipsum(self.ipsum).dolor(self.dolor)
}
```

This would be helpful in structs with many attributes. Additionally, it would allow to use immutable structs and whenever we need to perform a change we create a new one. Yes, it might not be the perfect approach, but in some contexts it might make sense.

TedDriggs:
Ah, I misunderstood. I think the more idiomatic way to express that in Rust would be `impl From<Lorem> for LoremBuilder { ... }`, but I've never seen that in any other crate's builders - or in the builders exposed by the standard library. Can you point to some examples of this in other Rust code please?

mseravalli:
[...]

Issue #73, opened 2017-04-11
Optionally expose a static Target::builder() -> TargetBuilder method


TedDriggs:
Proposal: On the type where `#[derive(Builder)]` is present, generate an inherent `impl` block for that type which defines a new `builder` method. This has a couple usability perks:

1. The caller doesn't have to import the type and the builder separately.
1. The call is shorter: `RequestOptions::builder()` vs. `RequestOptionsBuilder::default()`

colin-kiegel:
Interesting proposal. :-)
This is closely related to generating a customizable constructor `FooBuilder::new` #69.
cc @faern

Important questions are:
- would the behaviour be the same as in #69?
  - is it opt-in or opt-out?
  - should we offer `skip` / `name` / `private` attributes?
- would it exist in parallel with `FooBuilder::new`? How should we then namespace the settings?
  - `#[builder(constructor(static))]` vs `#[builder(constructor)]`?
    E.g. skipping both `#[builder(constructor(static(skip), skip))]`
  - `#[builder(static_constructor)]` vs `#[builder(constructor)]`?
    E.g. skipping both `#[builder(static_constructor(skip), constructor(skip))]`

A general note on both constructor proposals: If a user of `derive_builder` adds a default value to a previously required field, that would change the signature of every constructor which takes required fields in their signature. I.e. _adding_ default values would be a breaking change for every user of `derive_builder`. We should give a clear hint in the documentation about that and encourage library authors to write short unit tests for their builder API.

colin-kiegel:
My spontaneous tendency would be
- offer both, with namespaced settings like `#[builder(static_constructor)]` and `#[builder(constructor)]`
- make both opt-**out** (i.e. on by default)

faern:
Where does the name `static_constructor` come from? All constructor are static methods on a type. `FooBuilder::default()` is also a static method / constructor.

What need does this `Target::builder()` method fulfill that is currently not supported? To me it sounds like added sugar. Personally I think it's a higher priority to nail the design with infallible build + `TargetBuilder::new` constructor + structured errors + custom validation, which all add important features and might be hard to design nicely.

Con: Occupies a method name on the target type, we otherwise don't touch the target type(?)

colin-kiegel:
[...]

Issue #69, opened 2017-04-09
generate a (customizable) constructor


colin-kiegel:
Currently all builder structs implement the `Default` trait. Using the `Default` trait is the proposed way to create a new builder instance right now.

As mentioned in [this comment](https://github.com/colin-kiegel/rust-derive-builder/issues/56#issuecomment-292799381) by @faern it would be nice to have a dedicated constructor. We could make the identifier customizable (defaulting to `new`) and we could offer opting out of generating a constructor.

Possible API
- `#[builder(constructor(name="brand_new", ...))]`
- `#[builder(new(name="foo", private, ...))]`
- ...

Questions:
- should we offer renaming the constructor?
- should we offer to make the constructor private? (such that it can be wrapped - for easier re-ordering, or 
 something else)?
- how should we namespace these settings?

# Outlook

This could allow better support for infallible build methods #56, if the build method would already take all required values (`new(a, b)` instead of `new().a(a).b(b)`).

Limitations:
- the order of required fields in the constructor signature would most likely be identical with the order of fields. I would not like to offer an API for reordering fields, or something similar. In certain edge cases you might need implement the constructor manually - which will always be possible of course.

faern:
I like the `#[builder(constructor(...))]` approach better than `#[builder(new(...))]`, since constructor is a more correct name for the type of function, whereas "new" usually is the name of the function.

I think starting out simple can be desirable, not possible to rename the constructor and not possible to reorder fields.

Are you thinking that if one ask to get a constructor generated for them, it automatically takes all required fields as arguments, or would the default be to just work like how `default` works today?

colin-kiegel:
Ok, I also like `#[builder(constructor(...))]`, but just wanted to ask the question, especially since it's a bit longish.

Yes, starting simple is a valid option. But I don't think I would ever support reordering fields - just wanted to make that clear upfront.

Yes IMO the constructor should always take all required fields (those that don't have a default). This behaviour would be completely independent of fallible/infallible choice. This has the advantage that method signatures are somewhat more consistent across different derived builders. The new function would always tell you the required fields. If you then opt into infallible build methods and you do have required fields, then the `Default` trait would not be implemented anymore and the signature of the build method would change.

Does that make sense? :-)

faern:
I agree with never supporting field reorders. Then one can just reorder the fields of their struct instead to get that result.

[...]

PR #265, opened 2022-10-19
Added builder() fn for the original struct which returns the actual builder

alexzanderr:
Hello.

Love your crate, thanks for the code.

Here I've added just a simple `builder()` function for the original struct, this is a small PR, but I really want that change in the original crate, to not make wrappers.

What I've changed in this PR:
- added `builder()` function for the original struct to create a new builder from the original struct. Example:
```rust
fn main() {
    // create a new `ChannelBuilder` using `builder` fn from the `Channel` struct
    let ch = Channel::builder()
        .special_info(42u8)
        .token(19_124)
        .build()
        .unwrap();
    println!("{:#?}", ch);
}
```
Note the `Channel::builder()`, wasn't before.
- added a rustfmt.toml, assuming that you didnt have formatting configured, mine is configured and I dont want to mess your code formatting
- added some things to ignore in .gitignore, cuz I'm using a Makefile and maybe you dont use
- added `builder_function_for_the_struct.rs` example to test the usage of the `builder()` fn

Additional notes:
- there is nothing to test, since its a simple wrapper function 

Thats it. I assuming that you won't like the code structure, it didnt follow your pattern, we'll see.
In the first place I tried to add the changes to the `to_tokens` function, but I realised that this function doesnt get `ast: syn::DeriveInput` as parameter and also `Builder` struct has no fields with `ast`, has everything except the name of the original struct. I really needed the name of the struct, the name of struct wasn't defined in the `to_tokens` function, so I made a separate method for `Builder` to call from the macro.

Anyway, I'm open to code suggestions if you dont like it.

Thanks.

### DIFF
--- a/derive_builder/examples/builder_function_for_the_struct.rs
+++ b/derive_builder/examples/builder_function_for_the_struct.rs
@@ -1,0 +1,24 @@
+// NOTE: generate fully expanded version with `cargo expand`.
+//
+//       cargo expand --example new_builder_function
+#![allow(dead_code)]
+
+#[macro_use]
+extern crate derive_builder;
+
+#[derive(Default, Builder, Debug)]
+#[builder(setter(into))]
+struct Channel {
+    token: i32,
+    special_info: i32,
+}
+
+fn main() {
+    // create a new `ChannelBuilder` using `builder` fn from the `Channel` struct
+    let ch = Channel::builder()
+        .special_info(42u8)
+        .token(19_124)
+        .build()
+        .unwrap();
+    println!("{:#?}", ch);
+}
--- a/derive_builder_core/src/builder.rs
+++ b/derive_builder_core/src/builder.rs
@@ -140,6 +140,48 @@ pub struct Builder<'a> {
     pub std: bool,
 }
 
+impl<'a> Builder<'a> {
+    #[doc = "
+        adding builder() fn on the original struct
+        ```rust
+        impl <original_struct> {
+         pub fn builder() -> <original_struct>Builder  {
+             <original_struct>Builder::default()
+         }
+        }
+        ```
+        Why ?
+        Because sometimes @alexzanderr likes to call the builder like this:
+        ```rust
+            let ch = Channel::builder()
+            .special_info(42u8)
+            .token(19_124)
+            .build()
+            .unwrap();
+        ```
+        Note the `Channel::builder()`, it generates the builder from the original struct.
+    "]
+    pub fn add_builder_fn_to_struct(&self, ast: &syn::DeriveInput) -> TokenStream {
+        // name of the original struct
+        let struct_name_ident = &ast.ident;
+        // name of the builder for the original struct
+        let builder_name_ident = &self.ident;
+
+        // rust source code
+        let expanded = quote! {
+            /// This returns a new builder for the original struct
+            impl #struct_name_ident {
+                pub fn builder() -> #builder_name_ident {
+                    #builder_name_ident::default()
+                }
+            }
+        };
+
+        // this token stream is appended in in `builder_for_struct`
+        TokenStream::from(expanded)
+    }
+}
+
 impl<'a> ToTokens for Builder<'a> {
     fn to_tokens(&self, tokens: &mut TokenStream) {
         if self.enabled {
--- a/derive_builder_core/src/lib.rs
+++ b/derive_builder_core/src/lib.rs
@@ -86,5 +86,8 @@ pub fn builder_for_struct(ast: syn::DeriveInput) -> proc_macro2::TokenStream {
 
     builder.push_build_fn(build_fn);
 
-    quote!(#builder)
+    // adds pub fn builder() to original struct
+    let extra_builder_fn = builder.add_builder_fn_to_struct(&ast);
+
+    quote!(#builder #extra_builder_fn)
 }
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 /rls
 target
 Cargo.lock
+
+Makefile


derive_builder_core/src/lib.rs
```diff
@@ -86,5 +86,8 @@ pub fn builder_for_struct(ast: syn::DeriveInput) -> proc_macro2::TokenStream {
 
     builder.push_build_fn(build_fn);
 
-    quote!(#builder)
+    // adds pub fn builder() to original struct
+    let extra_builder_fn = builder.add_builder_fn_to_struct(&ast);
+
+    quote!(#builder #extra_builder_fn)
```
TedDriggs, line 92: This absolutely must not be added as on-by-default; doing so would break any existing code that manually added an inherent method called `builder`.

There would need to be a new field on `macro_options::darling_opts::Builder` to opt into this behavior; either a `Flag`, or an `Option<Ident>` to allow overriding the name would suffice.

That added field would also be cross-checked against the `custom_constructor` field to produce a precise and helpful error message if the two had incompatible settings (this is described in more detail in another review comment)

derive_builder_core/src/builder.rs
```diff
@@ -140,6 +140,48 @@ pub struct Builder<'a> {
...
+        let struct_name_ident = &ast.ident;
+        // name of the builder for the original struct
+        let builder_name_ident = &self.ident;
+
+        // rust source code
+        let expanded = quote! {
+            /// This returns a new builder for the original struct
+            impl #struct_name_ident {
+                pub fn builder() -> #builder_name_ident {
+                    #builder_name_ident::default()
```
TedDriggs, line 175: This should be using the standard from elsewhere in the crate of addressing the `default` method; see [this line](https://github.com/colin-kiegel/rust-derive-builder/blob/master/derive_builder_core/src/default_expression.rs#L33) for the right way to do that.

If this isn't done, a builder struct with an inherent method named `default` would run the wrong method here, as explained in The Book [here](https://web.mit.edu/rust-lang_v1.25/arch/amd64_ubuntu1404/share/doc/rust/html/book/first-edition/ufcs.html).

derive_builder/examples/builder_function_for_the_struct.rs
```diff
@@ -0,0 +1,24 @@
...
+#[derive(Default, Builder, Debug)]
+#[builder(setter(into))]
+struct Channel {
+    token: i32,
+    special_info: i32,
+}
+
+fn main() {
+    // create a new `ChannelBuilder` using `builder` fn from the `Channel` struct
+    let ch = Channel::builder()
```
TedDriggs, line 18: Thanks for including an example with this!

TedDriggs (changes requested):
Thank you for submitting a PR - `derive_builder` relies on its community to keep it going, and I appreciate the time you put into this.

The proposal of adding a method to the deriving struct has been discussed before, first in #73, then in #170, and again in #223. At the time, I noted in [this comment](https://github.com/colin-kiegel/rust-derive-builder/issues/170#issuecomment-665087279) that there are good reasons to want such a method - it makes `use` statements cleaner, improves builder discoverability, and makes the final code read more nicely.

That said, I felt then and still feel today that there is a stronger case _against_ `derive_builder` generating such a method itself.

* It messes up the documentation of the deriving struct by adding another `impl` block
* The code savings are very small, since the body of the method never needs to change

Since my original message, the increased interest in infallible builders through `custom_constructor` adds another reason not to implement this. We decided in #69 not to generate custom constructors which take arguments, with the stated workaround being to hand-write the constructor method which then feeds into a builder. I think the rationale there holds true here.

I'm inclined to close this PR, but since you put the time in to write it, I have still given it a full review. If we change our mind on this in the future, your work will provide a blueprint for how we might go about implementing this.

alexzanderr:
Thanks a lot for the response!

I will try to apply the all of the above suggestions and I will come back with updates on this PR or a make new PR.

derive_builder/examples/builder_function_for_the_struct.rs
alexzanderr: at least thats the only thing good in this PR 😅

TedDriggs:
> I will try to apply the all of the above suggestions and I will come back with updates on this PR or a make new PR.

Please note that addressing the review comments won't address the larger objections raised [here](https://github.com/colin-kiegel/rust-derive-builder/pull/265#pullrequestreview-1147489685) and in #170, so I still wouldn't anticipate this functionality being merged.

alexzanderr:
Sure I see your point.

But if I fix all the review problems and try to make this work without any problems and without any messes, would you approve it?

I mean, there is a big IF.

But woulnd't you approve it if hypothetically I can make this work?

TedDriggs:
> But if I fix all the review problems and try to make this work without any problems and without any messes, would you approve it?

Unlikely. The problems noted in #73 aren't something that _any_ implementation can fix; there's simply no way to make this an option of the derive-macro and avoid this method being in a second `impl` block, which messes up the docs for the deriving type. And - as discussed in #170 - it's not clear what we'd name the flag for something like this.

derive_builder_core/src/builder.rs
```diff
@@ -140,6 +140,48 @@ pub struct Builder<'a> {
...
+                }
+            }
+        };
+
+        // this token stream is appended in in `builder_for_struct`
```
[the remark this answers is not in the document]
alexzanderr: you are saying that rust-analyzer will add those comments alongside your actual references objects when running `find all refs` command inside your code editor/IDE.

so for the user developer those comments will be garbadge data? right?

question: is that why in all serious rust crates I've never seen a single comment? and also thats why your crate doesnt have any comments?

if so: but man, we need to document our code, how do you know what it does if you come over the code after 1 year without scratching your head off?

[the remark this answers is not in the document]
TedDriggs: I can't speak to why every crate behaves a certain way, and there are plenty of examples of comments in this crate ([example](https://github.com/colin-kiegel/rust-derive-builder/blob/master/derive_builder_core/src/initializer.rs)).

My tendency is to document the _outside_ of something if - when I come back to the code after some time away - I find myself confused about what the body of that thing is doing. That approach tends to have two positive effects:
1. I know which pieces of the code need further explanation; this can be hard to determine immediately after writing the code
2. The comment is unlikely to become stale without the person changing the file seeing the comment. Stale comments are worse than no comments because they can cause people to question the comments they do see in the codebase.

PR #275, opened 2022-11-22
Add #[builder(crate = "...")] for re-export scenarios

### LINKED ISSUE #274: Support re-export of derive_builder
## Background

[parol](https://github.com/jsinger67/parol) generates code that has several external crates, including derive_builder. And, [parol_runtime](https://github.com/jsinger67/parol_runtime) re-exports the crates for ease of use.

## Problem

I know `derive_builder` doesn't support crate renaming, but it also seems not to support re-exporting.
So we need a user-confusing workaround in `lib.rs` or `main.rs` as below:

```rust
use parol_runtime::derive_builder;
```

## Solution

Replace `::derive_builder" with "derive_builder" in generated code.

It may break projects in Rust 2015 edition, and we may need a feature toggle for this behavior.

TedDriggs:
This branch adds a new struct-level field which overrides the default crate path. This enables crates to re-export `derive_builder::Builder` and avoid any direct dependency on `derive_builder` in their downstreams.

Fixes #274


TedDriggs:
The compiletest error here is legitimate; we seem to have lost a span on the custom error in this change, and I'm not sure where yet. That needs to be addressed before this can be merged.

derive_builder_core/src/initializer.rs
```diff
@@ -155,54 +166,66 @@ pub enum FieldConversion<'a> {
...
+            MatchNone::ReturnError {
+                ref field_name,
+                ref span,
+                crate_root,
+            } => {
                 let conv_span = span.unwrap_or_else(Span::call_site);
-                let err_conv = quote_spanned!(conv_span => ::derive_builder::export::core::convert::Into::into(
-                    ::derive_builder::UninitializedFieldError::from(#field_name)
+                let err_conv = quote_spanned!(conv_span => #crate_root::export::core::convert::Into::into(
+                    #crate_root::UninitializedFieldError::from(#field_name)
```
TedDriggs, line 205: I'm not sure why yet, but this seems to be the problem. Changing `#crate_root` back to a hardcoded value fixes the error span that's failing the compiletest.

TedDriggs:
@ryo33 can you test this out using a git dependency and let me know how it goes?

ryo33:
@TedDriggs All worked well. Thank you!

PR #297, opened 2023-08-23
bump pretty_assertions to 1.2.1

alexanderkjall:
ref: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/master/src/derive-builder-core/debian/patches/relax-dev-deps.diff


TedDriggs:
Can you try rebasing this? I think that will fix the one CI failure, and I can then get this merged.

alexanderkjall:
done :)

PR #302, opened 2023-12-22
Allows prevention of `alloc` dependency through feature and disabling of `ValidationError(String)` variant.

red1bluelost:
Currently, using `derive_builder` in `no_std` still forces the user to make a `global_allocator` since the crate links against `alloc`. This PR tries to avoid that in two aways.
* Create a new feature flag `alloc` that controls whether the `alloc` crate is used. This allows `no_std` user to opt into using allocation if they want it.
* New `build_fn` field, `validation_error: bool` that controls whether the generated error includes `ValidationError(String)`. This prevents usage of any `alloc` code but also limits the user from using the `validate` field.
  * The user can provide a custom `build_fn` `error` to avoid `alloc` code from being used.

This is to fix issue: https://github.com/colin-kiegel/rust-derive-builder/issues/303.

### DIFF
--- a/derive_builder_core/src/builder.rs
+++ b/derive_builder_core/src/builder.rs
@@ -132,8 +132,8 @@ pub struct Builder<'a> {
     /// Whether to include `ValidationError` in the generated enum. Necessary to avoid dependency
     /// on `alloc::string`.
     ///
-    /// This would be `false` when `build_fn.validation_error == false`. This is ignored when
-    /// `generate_error` is `false`.
+    /// This would be `false` when `build_fn.error.as_validation_error() == Some((false, _))`. This
+    /// has no effect when `generate_error` is `false`.
     pub generate_validation_error: bool,
     /// Whether this builder must derive `Clone`.
     ///
--- a/derive_builder_core/src/builder.rs
+++ b/derive_builder_core/src/builder.rs
@@ -12,6 +12,11 @@ use BuilderPattern;
 use DeprecationNotes;
 use Setter;
 
+const ALLOC_NOT_ENABLED_ERROR: &str = r#"`alloc` is disabled within 'derive_builder', consider one of the following:
+* enable feature `alloc` on 'dervie_builder' if a `global_allocator` is present
+* use a custom error `#[builder(build_fn(error = "path::to::Error"))]
+* disable the validation error `#[builder(build_fn(error(validation_error = false)))]"#;
+
 /// Builder, implementing `quote::ToTokens`.
 ///
 /// # Examples
@@ -135,6 +140,8 @@ pub struct Builder<'a> {
     /// This would be `false` when `build_fn.error.as_validation_error() == Some((false, _))`. This
     /// has no effect when `generate_error` is `false`.
     pub generate_validation_error: bool,
+    /// Indicator of `cfg!(not(any(feature = "alloc", feature = "std")))`, as a field for tests
+    pub no_alloc: bool,
     /// Whether this builder must derive `Clone`.
     ///
     /// This is true even for a builder using the `owned` pattern if there is a field whose setter
@@ -233,7 +240,10 @@ impl<'a> ToTokens for Builder<'a> {
                 ));
             }
 
-            if self.generate_error {
+            if self.no_alloc && self.generate_error && self.generate_validation_error {
+                let err = syn::Error::new_spanned(&self.ident, ALLOC_NOT_ENABLED_ERROR);
+                tokens.append_all(err.to_compile_error());
+            } else if self.generate_error {
                 let builder_error_ident = format_ident!("{}Error", builder_ident);
                 let builder_error_doc = format!("Error type for {}", builder_ident);
 
@@ -385,6 +395,7 @@ macro_rules! default_builder {
             functions: vec![quote!(fn bar() -> { unimplemented!() })],
             generate_error: true,
             generate_validation_error: true,
+            no_alloc: false,
             must_derive_clone: true,
             doc_comment: None,
             deprecation_notes: DeprecationNotes::default(),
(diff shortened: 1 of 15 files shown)


derive_builder_core/src/builder.rs
```diff
@@ -231,15 +237,35 @@ impl<'a> ToTokens for Builder<'a> {
                 let builder_error_ident = format_ident!("{}Error", builder_ident);
                 let builder_error_doc = format!("Error type for {}", builder_ident);
 
+                let validation_error = self
+                    .generate_validation_error
+                    .then_some(quote!(
```
TedDriggs, line 242: This is pretty cool, and I hadn't seen it before. However, if the discussion in #303 does lead to this being on track for a merge, we need to avoid using a method that isn't available to 1.56.0; can you change each use of it to an `if` block?

red1bluelost: Updated the code then built and ran cargo test using 1.56.0 x86_64.

TedDriggs:
Apologies; I didn’t get a chance to review this today, so it’ll probably be about a week before I next get the chance to. I do intend to review it, and think we are on track towards merging this

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -95,6 +95,25 @@ pub struct BuildFn {
     /// * If `validate` is specified, then this type must provide a conversion from the specified
     ///   function's error type.
     error: Option<Path>,
```
TedDriggs, line 97: I'm wondering if we should make this into a type that can either be a key-value pair with a path, or a list that accepts options about the generated error:


`error = "path"`


`error(validation = false)`

The advantage would be that it's then impossible to incorrectly disable the validation variant when using a custom error.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -95,6 +95,25 @@ pub struct BuildFn {
...
+    validation_error: bool,
+}
+
+impl BuildFn {
+    fn validation_needs_error(self) -> darling::Result<Self> {
+        if self.validate.is_some() && !self.validation_error {
+            Err(darling::Error::custom(
+                "Cannot set `validation_error = false` when using `validate`",
+            )
+            .with_span(&self.validate))
```
TedDriggs, line 112: The wording of the error message feels incorrectly paired with its span; I would either make the message be "Cannot use `validate` when `validate_error = false`" or I would set the span to be the `validate_error` field.

derive_builder_core/src/macro_options/darling_opts.rs
```diff
@@ -95,6 +95,25 @@ pub struct BuildFn {
...
+    /// Whether to generate `ValidationError(String)` as a variant of the build error type.
+    ///
+    /// Setting this to `false` will prevent `derive_builder` from using the `validate` function
+    /// but this also means it does not generate any usage of the `alloc` crate (useful when
+    /// disabling the `alloc` feature in `no_std`).
+    validation_error: bool,
+}
+
+impl BuildFn {
+    fn validation_needs_error(self) -> darling::Result<Self> {
```
TedDriggs, line 107: We should make sure someone doesn't use both `validation_error` and `error` in this method too. (And use `darling::Error::accumulator` to make sure we get all the errors gathered up at once before returning)

TedDriggs (commented):
Did my best to review from mobile

derive_builder_core/src/macro_options/darling_opts.rs
red1bluelost: I managed to implement it so it supports the two forms from your comment.

derive_builder_core/src/macro_options/darling_opts.rs
red1bluelost: Agreed, I changed the span to point to `validation_error` now.

derive_builder_core/src/macro_options/darling_opts.rs
red1bluelost: I moved to the accumulator in case future stuff wants to use this method. `error` and `validation_error` is now impossible via the parse since I'm using an enum for it (from your other comment).

red1bluelost:
Thank you for the review. No worry about time, I'm in not in a hurry for this feature.

I'm looking into the workflow breakage. This PR has somewhat of a breaking change since `alloc` is an additive feature for something that was previously always enabled. For the workflow, `--no-default-features` before this PR is similar to `--no-default-features --features alloc` now.

red1bluelost:
My most recent commit addresses the GH workflows. It adds `alloc` feature on `--no-default-features` to get the same no_std with alloc behavior as before.

I also added a compile error when `std` and `alloc` are disabled but a user attempts to derive with `alloc::String` usage. It gives suggestion to enable the `alloc` feature or avoid String by using custom error or disabling validation error. This should help no std users update there code when upgrading.

TedDriggs:
I just pushed a simplified version of the `BuildFnError` code [here](https://github.com/colin-kiegel/rust-derive-builder/tree/no-alloc), but I'm not sure why it isn't showing up in the PR.

red1bluelost:
> I just pushed a simplified version of the `BuildFnError` code [here](https://github.com/colin-kiegel/rust-derive-builder/tree/no-alloc), but I'm not sure why it isn't showing up in the PR.

I think it's because this PR is on my fork while your code is a branch here. I pulled the commit and just pushed it.

Thanks for simplifying the macro code! I had a feeling that there was a nicer way to do it but this was my first time using darling_opt :)

TedDriggs:
Please rebase this; I think that will fix the CI issue. Once we have a clean bill of health from CI, we are ready to merge.

red1bluelost:
Rebased it :+1:

PR #306, opened 2024-02-01
Accept field(ty = "...") to ease migration to syn 2.0

TedDriggs:
On syn 2.0, field(type = "...") will break with a not-very-informative error message.

While we can't prevent that, we can make it possible for people to switch to the new property name without needing to change anything else, so that the eventual dependency version bump is fully source-compatible.

# Alternatives Considered
I debated using `DeprecationNotes` to put a warning on all uses of `type`, but decided it wasn't worth the code churn necessary to plumb that through from the options area all the way to code generation. This is already a highly specialized feature, and the fix is pretty easy once you read the release notes.

### DIFF
--- a/derive_builder/CHANGELOG.md
+++ b/derive_builder/CHANGELOG.md
@@ -2,6 +2,9 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## Unreleased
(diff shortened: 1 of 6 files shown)


jtran (approved):
I don't have much experience using `darling`. But I read the code and tried it on my codebase, and it looks good to me.

PR #308, opened 2024-02-12
Update to syn 2 and darling 0.20.5

### LINKED ISSUE #292: Update syn dependency?
Hello,

Thank you for making this project.

I was wondering if there was any plan to upgrade `syn` to v2 anytime soon.  My motivation is that I have multiple versions of `syn` in my dependency tree, and I'm trying to get it down to one to reduce build times.

TedDriggs:
This uses the most recent major version of `syn`, as well as the most recent version of `darling`.

# Breaking Changes
1. It's no longer possible to use `type` to override the type of a field; instead, `ty` must be used.

Fixes #292

### DIFF
--- a/derive_builder/Cargo.toml
+++ b/derive_builder/Cargo.toml
@@ -7,6 +7,7 @@ authors = [
     "Jan-Erik Rediger <janerik@fnordig.de>",
     "Ted Driggs <ted.driggs@outlook.com>",
 ]
+edition = "2018"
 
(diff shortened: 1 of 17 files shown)


jtran (approved):
Looks good to me. I built it my project with this branch, and it looks good.

TedDriggs:
I'm going to wait to merge this until TedDriggs/darling#263 has landed, as I may want to ship with a darling dependency of 0.20.6 instead of 0.20.5. However, that hopefully won't take long to resolve, and at that point this will be unblocked.

Issue #298, opened 2023-10-23
Owned pattern does not compile with custom default function.


Wasabi375:
```rust
struct V(u32);

#[derive(Builder)]
#[builder(pattern = "owned")]
struct Foo {
    a: V,
    #[builder(default = "self.calc_b()")]
    b: V,
}

impl FooBuilder {
    fn calc_b(&self) -> V {
        V(self.a.as_ref().unwrap().0 + 5)
    }
}
```
This fails to compile with
```
borrow of partially moved value: `self`
 --> src/lib.rs:9:25
  |
3 | #[derive(Builder)]
  |          ------- value partially moved here
...
7 |     #[builder(default = "self.calc_b()")]
  |                         ^^^^^^^^^^^^^^^ value borrowed here after partial move
```

Looking at cargo expand it's easy to see why:
```rust
fn build(
        self,
    ) -> ::derive_builder::export::core::result::Result<Foo, FooBuilderError> {
        Ok(Foo {
            a: match self.a {
                // partial move here
                Some(value) => value,
                None => {
                    return ::derive_builder::export::core::result::Result::Err(
                        ::derive_builder::export::core::convert::Into::into(
                            ::derive_builder::UninitializedFieldError::from("a"),
                        ),
                    );
                }
            },
            b: match self.b {
                Some(value) => value,
                // borrow after partial move here
                None => self.calc_b(),
            },
        })
    }
```
A possible fix would be to calculate custom default values before the match expression.
A small issue with that is, that the default value will be calculated even if not needed, but I don't think that is a big problem here and can be ignored, except for maybe mentioning it in the documentation.

I'd be willing to create a PR for this if this solution is acceptable.

Wasabi375:
Let me think about this some more. There should be a way to implement this without calculating all default values.

Would you be willing to accept a PR if I can implement this as a zero-cost abstraction?

TedDriggs:
In principle, yes. However, I’m skeptical that it’s possible to do this without either cloning `self.a` or computing the default when you don’t technically need to.

Wasabi375:
I think I found a way to do this without cloning or computing the default value when not necessary.
The generated code should look something like this at the end:
```rust
fn build(self) -> ::derive_builder::export::core::result::Result<Foo, FooBuilderError> {
    let default_a: Option<V> = match self.a.as_ref() {
        Some(_) => None,
        None => {
            return ::derive_builder::export::core::result::Result::Err(
[...]

PR #315, opened 2024-03-24
Support for custom default fuction with owned pattern

Wasabi375:
This is a fix for #298

I finally got back to my hobby project where I need this so I got around to implement this feature/bug fix.

I've implemented the solution I described in that issue so I won't repeat it here again. If there are any open questions I'm happy to answer them or adjust this PR.

> You wouldn’t need that unwrap if you use unwrap_or, right?

No unwrap_or does not work because it expects a `T` and the default we have is an Option. Either we use or and than unwrap or we can use `unwrap_or(default.unwrap())`. I personally think that `or.unwrap` is a bit nicer. 

I know that the rust 1.59 tests are failing right now. "Syn" just published a new version "2.0.55" that requires rust 1.60.
However on my fork the 1.59 tests succeeded with the old syn version.

### DIFF
--- a/derive_builder/CHANGELOG.md
+++ b/derive_builder/CHANGELOG.md
@@ -2,6 +2,9 @@
 All notable changes to this project will be documented in this file.
 This project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+- Allow default values that access `self` in combination with the "owned" pattern. #298
+
 ## [0.20.0] - 2024-02-14
 - Bump `syn` to version 2 #308
 - Bump `darling` to version 0.20.6 #308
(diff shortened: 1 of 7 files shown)


Wasabi375:
Woops. Turns out my simplified testcase isn't actually good enough, so I will have to do some more work on this. I'll leave this as a draft PR for now.

Wasabi375:
Ok, I spent the last month making sure that I did not miss anything and properly testing, this. So it should be good to merge. 
Again let me know if there are any changes you'd like me to make or if I should rebase the git history to clean it up (otherwise squashing this should be fine I think).

TedDriggs:
I'm not sure when I'll have time to review this, but I have it on my to-do list.

Issue #282, opened 2023-02-20
Copy default generic types from source


ColonelThirtyTwo:
Example code:

```rust
use derive_builder::Builder;


#[derive(Builder)]
pub struct Foo<T = u32> {
    pub v: T
}
```

The derived `FooBuilder` has a single generic type `<T>` like `Foo`, but it does not copy the default type `u32` from `Foo` - it must be specified manually or inferred. It would be convenient if the builder copied the default type.

TedDriggs:
This is likely because of [this line](https://github.com/colin-kiegel/rust-derive-builder/blob/master/derive_builder_core/src/builder.rs#L153); we use the `syn` method for splitting generics for codegen, but that assumes the caller is creating a new `impl` block rather than redeclaring the struct.

If someone can submit a PR with it and some tests, then I'll be happy to review it - otherwise, I'll take a look when I'm next able.

TedDriggs:
@ColonelThirtyTwo I tried changing that line, and successfully got the defaults to emit. However, I'd have expected the change to make the following code work, and it doesn't:

```rust
#[derive(Debug, PartialEq, Default, Builder, Clone)]
pub struct GenericWithDefault<T = usize> {
    #[builder(default, setter(strip_option))]
    lorem: Option<T>,
}

#[test]
fn generic_builder_with_defaults() {
    let x = GenericWithDefaultBuilder::default();
    let y = x.build().unwrap();

    assert_eq!(y, GenericWithDefault { lorem: None });
}
```

The compiler says it cannot figure out the type of `x`, which suggests to me that the default value for `T` doesn't flow through the usage of `default()`. If that is the case, what's the value of this change?

ColonelThirtyTwo:
I'm not sure what you'd like me to test without seeing the code that you changed.

My particular use case is more along the lines of:

```rust
#[derive(Builder)]
#[builder(pattern = "owned")]
pub struct Options<Error = std::io::Error> {
    pub on_error: Box<dyn FnMut(Error)>;
}
```

Ideally error would default to `std::io::Error`. Maybe rust's type inference isn't good enough to suss that out, but I can't test it without the change you mae.

TedDriggs:
It's on the branch [generics](https://github.com/colin-kiegel/rust-derive-builder/tree/generics)

TedDriggs:
@ColonelThirtyTwo did the branch fix your issue?

PR #319, opened 2024-05-09
forward default typeparam

tharvik:
the derived builder doesn't offer the same default types as the source struct. that forces the user to explicit all types when theses can't be infered.

in the case I have, I want to derive a builder on a struct with a bunch of callbacks:
```rust
pub struct Settings<T, S = fn(T), R = fn(T), C = fn(T)>
where S: FnOnce(T), R: FnMut(T), C: FnMut(T)
```
on which I only need to specific `T` when instanciating `Settings` "by hand". but when using the derived builder, I need to specifiy *all* the types (which is especially cumbersome as most of theses callbacks are optional).

there was a confusion in the implementation between generics meant for the builder struct (which [can contain default types](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#default-generic-type-parameters-and-operator-overloading)) and the ones for `impl` (which [can't contain theses](https://github.com/rust-lang/rust/issues/36887)). this PR fixes that by directly using the generics specified on the struct (with bounds and defaults), and a test to guard against regression for this kinda obscure feature.

I encountered dtolnay/syn#782 while doing that but as it seems that upstream seems to have forgot this issue, a workaround seemed better.

### DIFF
--- a/derive_builder/src/lib.rs
+++ b/derive_builder/src/lib.rs
@@ -667,7 +667,7 @@
 //! - Tuple structs and unit structs are not supported as they have no field
 //!   names.
 //! - Generic setters introduce a type parameter `VALUE: Into<_>`. Therefore you can't use
-//!  `VALUE` as a type parameter on a generic struct in combination with generic setters.
+//!   `VALUE` as a type parameter on a generic struct in combination with generic setters.
 //! - The `try_setter` attribute and `owned` builder pattern are not compatible in practice;
 //!   an error during building will consume the builder, making it impossible to continue
 //!   construction.
--- a/derive_builder/tests/run-pass/default_typeparams.rs
+++ b/derive_builder/tests/run-pass/default_typeparams.rs
@@ -1,0 +1,30 @@
+//! Test ensuring that default type params gets forwarded to Builder.
+
+#[macro_use]
+extern crate derive_builder;
+
+#[derive(Builder)]
+#[builder(setter(strip_option))]
+struct Settings<T, U = (), C = fn(T) -> U> {
+    first: T,
+    #[builder(default)]
+    second: Option<U>,
+    #[builder(default)]
+    third: Option<C>,
+}
+
+fn main() {
+    SettingsBuilder::<usize>::default()
+        .first(1)
+        .second(())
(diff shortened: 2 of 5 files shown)


TedDriggs:
When I last tried this in #282, I was surprised to find that it didn't work the way I'd predicted. Does the test case from that conversation work with this PR?

tharvik:
> When I last tried this in #282, I was surprised to find that it didn't work the way I'd predicted. Does the test case from that conversation work with this PR?

No, it doesn't, but that's a way larger issue on [how default type params are handled](https://github.com/rust-lang/rust/issues/98931). It won't ~work as predicted until upstream addresses it: one still needs to specify some generic params. but that reduces the number of needed params from my use case from four to one which is already a step in the right direction.

Note that slightly modified version of your test case would work but is quite weird to use
```rust
#[test]
fn generic_builder_with_defaults() {
    let x: GenericWithDefaultBuilder = Default::default();
    let y = x.build().unwrap();

    assert_eq!(y, GenericWithDefault { lorem: None });
}
```

tharvik:
@TedDriggs anyway to help in moving this along?

TedDriggs:
@tharvik I missed your previous reply. Thanks for doing the research into the upstream issue here. My one ask would be that you add a test to `derive_builder/tests/run-pass` which exercises your use-case, so we have an end-to-end demonstration of it working. Once that's added, we can merge this and I'll ship a new version of the crate.

tharvik:
great then, thanks for taking the time of handling this!
I added a small use-case at `derive_builder/tests/run-pass/default_typeparams.rs` which tests most features that I need (my actual use-case is way larger yet works flawlessly with this patch).

tharvik:
soo, stable rust got updated to 1.80 since last CI build, which introduces [a new lint](https://rust-lang.github.io/rust-clippy/master/index.html#/doc_lazy_continuation) and changed [the name of one](https://rust-lang.github.io/rust-clippy/master/index.html#/disallowed_names). latest commit takes care of that, a bit out of scope IMO, feel free to drop it.

TedDriggs:
This is published in v0.20.1

Issue #325, opened 2024-06-27
support suppressing default `Clone` derivation


mathstuf:
I'd like to instead use `perfect_derive(Clone)` than `derive(Clone)`. I can do the former with `builder_struct_addr`, but this conflicts with the `derive(Clone)` added by default. A direct dependency on `perfect_derive` seems excessive, but an escape hatch to do `Clone` in a different way would be nice.

mathstuf:
Suppression is more reliable with `pattern = "owned"` instead.

PR #326, opened 2024-06-27
Add `suppress_derive_clone` struct attribute

mathstuf:
This attribute suppresses the default-generated `derive(Clone)`. This may be necessary where `derive(Clone)` copies constraints in an unsatisfiable way.

See: https://smallcultfollowing.com/babysteps//blog/2022/04/12/implied-bounds-and-perfect-derive/
See: rust-lang/rust#26925
Fixes: #325

---
I tried running the `compile-fail` tests but couldn't seem to get them to appear when using `RUSTFLAGS='--cfg compiletests'` even though it seemed like it was building the required dependencies. Let's see what GHA says…

### DIFF
--- a/derive_builder/src/lib.rs
+++ b/derive_builder/src/lib.rs
@@ -504,6 +504,11 @@
 //! }
 //! ```
 //!
+//! The `Builder` will also automatically `derive(Clone)` in some cases. This may not always be
+//! suitable if there are generic types involved. This may be suppressed using the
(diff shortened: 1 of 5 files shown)


mathstuf:
`clippy` seems to be unhappy about generated code; ignoring.

mathstuf:
I ran into issues actually using this. It was easier to instead just `pattern = "owned"`.

PR #333, opened 2024-11-15
Impl core::error::Error instead of std::error::Error

DeppLearning:
Hey there,

thanks for the great library! I noticed that the Error trait is not implemented for BuilderErrors when the `std` feature is inactive. Rust 1.81 stabilized core::error::Error, it would be nice if this crate could make use of it and lift the former restriction. Since, as implemented, this would be a breaking change, I'm wondering what the preferred way forward is. I can think of

- add a feature flag, like `core_error` that opts into using core::error::Error instead of the std version
- use something like https://crates.io/crates/rustversion/ to gate the core version behind a  `#[rustversion::since(1.81)]`
- others?

### DIFF
--- a/derive_builder_core/src/builder.rs
+++ b/derive_builder_core/src/builder.rs
@@ -145,6 +145,7 @@ pub struct Builder<'a> {
     /// Doc-comment of the builder struct.
     pub doc_comment: Option<syn::Attribute>,
     /// Whether or not a libstd is used.
+    #[allow(dead_code)]
     pub std: bool,
 }
 
@@ -294,11 +295,9 @@ impl<'a> ToTokens for Builder<'a> {
                     }
                 ));
 
-                if self.std {
-                    tokens.append_all(quote!(
-                        impl std::error::Error for #builder_error_ident {}
-                    ));
-                }
+                tokens.append_all(quote!(
+                    impl core::error::Error for #builder_error_ident {}
+                ));
             }
         }
     }


TedDriggs:
> add a feature flag, like core_error that opts into using core::error::Error instead of the std version

I think I'd rather bump the MSRV than use a cargo flag for it. If the only difference between 0.20.x and 0.21.0 is an MSRV bump for this purpose, upgrading will be very easy for those who can upgrade, and those who can't upgrade won't be missing out on anything.

> use something like https://crates.io/crates/rustversion/ to gate the core version behind a #[rustversion::since(1.81)]

I wouldn't consider a version that included `rustversion` in the _generated_ code, but would be willing to have the `derive_builder_core` crate take that dependency. Would the code in the macro be something like...

```rust
if rustversion::cfg!(since(1.81)) {
    // generate core::error::Error impl
}
```

derive_builder_core/src/builder.rs
```diff
@@ -294,11 +295,9 @@ impl<'a> ToTokens for Builder<'a> {
                     }
                 ));
 
-                if self.std {
-                    tokens.append_all(quote!(
-                        impl std::error::Error for #builder_error_ident {}
```
TedDriggs, line 299: In ≥1.81, is there some blanket impl on `std::error::Error` that means all `core::error::Error` impls are satisfactory? If not, do we need to emit both impls to avoid breaking compatibility?

PR #349, opened 2025-09-25
Bump darling to version 0.21.3

brandonweeks:


TedDriggs:
@brandonweeks which item from v0.21.3 are you wanting to bring over to derive builder? There are some unreleased changes in `darling` going out next week, so I'm inclined to wait and do a move straight to 0.21.4.

brandonweeks:
I was just attempting to eliminate extra copies of dependencies in our build, so waiting for `darling` 0.21.4 is totally fine.

PR #352, opened 2025-12-30
Fix typo in README about setter strip option

markstory:
Replace a `'` with ` `.


TedDriggs:
The 1.56.0 failure isn't this PR's fault, so I'm going to merge and will deal with the 1.56.0 issue later.

